PR #14, opened 2015-09-07
Worker logger

agwozdowski:


Rakefile
```diff
@@ -5,6 +5,14 @@ require 'polishgeeks-dev-tools'
 PolishGeeks::DevTools.setup do |config|
   config.brakeman = false
   config.haml_lint = false
+  config.empty_method_ignored = %w(
+    spec/lib/karafka/base_controller_spec.rb
+    spec/lib/karafka/connection/cluster_spec.rb
+    spec/lib/karafka/connection/consumer_spec.rb
+    spec/lib/karafka/connection/listener_spec.rb
```
mensfeld: maybe we should just return self for all the methods - excluding so many specs doesnt seem right

PR #57, opened 2016-01-25
Routes

mensfeld:
@pavlo-vavruk  @agwozdowski  have fun ;)


lib/karafka/connection/consumer.rb
```diff
@@ -3,22 +3,21 @@ module Connection
...
       # @param message [Poseidon::FetchedMessage] message that was fetched by poseidon
-      def consume(controller_class, message)
-        controller = Karafka::Routing::Router.new(
-          Message.new(controller_class.topic, message.value)
-        ).build
+      def consume(message)
+        controller = Karafka::Routing::Router.new(message.topic).build
+        # We wrap it around with our internal message format, so we don't pass around
+        # a raw Poseidon message
+        controller.params = Message.new(message.topic, message.value)
```
pavlo-vavruk: Seems, it is only one place where we use Message wrapper, isn't it?
I think it's enough to send `message` as a param and change `Message` class implementation.

lib/karafka/base_worker.rb
```diff
@@ -0,0 +1,44 @@
...
+    # @param params [Hash] params hash that we use to build Karafka params object
+    def perform(topic, params)
+      self.topic = topic
+      self.params = params
+      Karafka.monitor.notice(self.class, params: params)
+      controller.perform
+    end
+
+    # What action should be taken when perform method fails
+    # @param topic [String] Topic bthat we will use to route to a proper controller
```
pavlo-vavruk: `bthat` typo

lib/karafka/routing/route.rb
```diff
@@ -0,0 +1,73 @@
...
+    # - interchanger - What interchanger to encode/decode data do we want to use (optional)
+    class Route
+      # Only ASCII alphanumeric characters and underscore and dash are allowed in topics and groups
+      NAME_FORMAT = /\A(\w|\-)+\z/
+
+      # Options that we can set per each route
+      attr_writer :group, :topic, :controller, :worker, :parser, :interchanger
+
+      # This we can get "directly" because it does not have any details, etc
+      attr_reader :controller
```
pavlo-vavruk: Maybe it is better to have ` attr_accessor :controller` and get rid of `:controller` within attr_writers?

lib/karafka/routing/route.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      # Initializes default values for all the options that support defaults if their values are
+      # not yet specified. This is need to be done (cannot be lazy loaded on first use) because
+      # everywhere except Karafka server command, those would not be initialized on time - for
+      # example for Sidekiq
+      def build
```
pavlo-vavruk: Can we for better code reading have methods order as it is below? 
```ruby
  group
  worker
  parser
  interchanger
  self
```

Issue #63, opened 2016-01-29
Graceful shutdown with current offset state during data processing


mensfeld:
When we trigger graceful shutdown, we should not process the whole bulk, instead we should stop processing after current message and commit the offset in this place

PR #64, opened 2016-01-29
Commit

mensfeld:
@agwozdowski  @pavlo-vavruk  please review - it is a fix for #63 
also protects from kill -9 and loosing data then and other weird cases that might occur.
Significantly speeds up shutdown time (from 20-60 seconds on a full load to less than 1 second)

### DIFF
--- a/lib/karafka/cli.rb
+++ b/lib/karafka/cli.rb
@@ -33,9 +33,8 @@ module Karafka
 end
 
 # This is kinda trick - since we don't have a autoload and other magic stuff
-# like Rails does, so instead this method allows us to exist with exitcode 10
-# then our wrapping process (outside of irb) detects this, it will restart the
-# console with new code loaded
+# like Rails does, so instead this method allows us to replace currently running
+# console with a new one via Kernel.exec. It will start console with new code loaded
 # Yes we know that it is not turbofast, however it is turbo convinient and small
 #
 # Also - the KARAFKA_CONSOLE is used to detect that we're executing the irb session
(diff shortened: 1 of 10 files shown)


lib/karafka/cli.rb
```diff
@@ -47,7 +46,7 @@ def cli_commands
   # Reloads Karafka irb console session
   def reload!
     puts "Reloading...\n"
-    Kernel.exit 10
+    Kernel.exec "KARAFKA_CONSOLE=true bundle exec irb -r #{Karafka.boot_file}"
```
pavlo-vavruk: Can we use here ` Kernel.exec  Karafka::Cli::Console.command` ?

lib/karafka/connection/queue_consumer.rb
```diff
@@ -81,6 +83,27 @@ def target
...
+      # @param partition [Integer] Number of partition from which we were consuming messages
+      # @param last_processed_message [Poseidon::FetchedMessage] last message that we've
+      #   successfully processed
+      # @note By processed we mean passing it through the controller and scheduling it to be
+      #   handled by the worker. Then we can consider data to be "saved" since it is being stored
+      #   in Redis and will be handled in a background job
+      # @note Most of the time, this method will be scheduled on a last message that we've received
+      #   from Kafka, but there are some cases (graceful shutdown, restart) when we will stop
+      #   processing fetched messages, return last that we've managed to handle and exit
+      # @note If is also worth pointing out, that if we kill Karafka process during the data
```
pavlo-vavruk: `@note If`  => `@note It`

pavlo-vavruk:
@mensfeld reviewed

lib/karafka/cli.rb
mensfeld: yeah good catch - that was a purpose of that!

mensfeld:
@pavlo-vavruk  thanks! Added tweaks and once the CI is green, will merge

Issue #65, opened 2016-02-01
Example of NewRelic monitor is outdated


90yukke:
NewRelic monitor have stopped to work after recent update of karafka.

mensfeld:
That is true and it is because of the change what and when is logged in the Karafka logging engine (monitoring). Work in progress

mensfeld:
@90yukke could you verify that it works?

mensfeld:
@agwozdowski @pavlo-vavruk once you've merged #70 please close this issue

mensfeld:
I'll close it since it works when I've tested it. If you'll have any issues with this, I will reopen and we can continue investigating.

PR #70, opened 2016-03-20
#65

mensfeld:
@pavlo-vavruk  could you verify this as well please

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,14 +424,20 @@ end
 Here's a simple example of monitor that is used to handle events and errors logging into NewRelic. It will send metrics with information about amount of processed messages per topic and how many of them were scheduled to be performed async.
 
 ```ruby
+# NewRelic example monitor for Karafka
 class AppMonitor < Karafka::Monitor
+  # @param [Class] caller class for this notice
+  # @param [Hash] hash with options for this notice
   def notice(caller_class, options = {})
+    # Use default Karafka monitor logging
     super
-    action = :"notice_#{caller_label}"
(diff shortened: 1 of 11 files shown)


spec/lib/karafka/monitor_spec.rb
```diff
@@ -76,4 +76,31 @@
...
+      [
+        Karafka::Runner
+      ]
+    end
+
+    it 'expect to have proper classes on error' do
+      expect(subject.send(:caller_exceptions_map)[:error]).to eq error_callers
+    end
+
+    it 'expect to have proper classes on fataal' do
```
agwozdowski: typo fataal

README.md
```diff
@@ -439,18 +445,35 @@ class AppMonitor < Karafka::Monitor
...
+  # @param topic [String] topic name
+  def consume(topic)
+    record_count metric_key(topic, __method__)
   end
 
-  def notice_perform_async(caller_class, _options)
-    record_count metric_key(caller_class, __method__)
+  # Log that message for topic were scheduled to be performed async
+  # @param topic [String] topic name
+  def perform_async(topic)
```
pavlo-vavruk: Since now `perform_async`, `consume` and `perform` methods have the same body, maybe it is better to use aliases?
Or there the documentation matters? Anyway we can add documentation to aliases.

spec/lib/karafka/monitor_spec.rb
pavlo-vavruk: `fatal`

Issue #59, opened 2016-01-27
Support for zookeeper chroot


sokratisg:
Hello,

Does consumer support chroot for zookeeper connections?
I am asking because on a multi-environment setup and each Kafka broker should running under a different "/<chroot>" node when all Kafka clusters are using the same Zookeeper cluster.

I've tried adding the chroot in config.zookeeper_hosts array but Karafka seems to ignore the setting.
`config.zookeeper_hosts =  %w( 10.0.0.1:2181/kafka_prd 10.0.0.2:2181/kafka_prd 10.0.0.3:2181/kafka_prd  )`

Thank you.

mensfeld:
True - will fix that soon

mensfeld:
@sokratisg  could you try to specify the chroot only for the last connection:

```
config.zookeeper_hosts = %w( 10.0.0.1:2181 10.0.0.2:2181 10.0.0.3:2181/kafka_prd )
```

and check if it will help

mensfeld:
@sokratisg it works with format as above (at least for me). Closing this issue since no reply. Please reopen if it still occurs an provide more details.

PR #74, opened 2016-04-09
Add zookeeper_base_znode config value

dwillett:
Our zookeeper setup does not store the kafka broker data at the root znode. This adds a configuration setting to specify the base path

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ Karafka has following configuration options:
 | redis                  | true     | Hash              | Hash with Redis configuration options                                                       |
 | wait_timeout           | true     | Integer (Seconds) | How long do we wait for incoming messages on a single socket (topic)                        |
 | zookeeper_hosts        | true     | Array<String>     | Zookeeper server hosts                                                                      |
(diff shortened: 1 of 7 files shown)


mensfeld:
@dwillett  thank you for the  merge request - we will review it soon. Technically it was already possible to use namespaced zookeeper by specifying the chroot: please see #59 issue - there's a solution. However it seems that it is an unclear solution so I will review and merge it.

I would also suggest (you don't need to change it, I will) to do exactly what we did with redis (I will create an issue for that and to merge your changes with it):

```ruby
zookeeper = {
  hosts: [...],
  base_znode: 'something'
}

kafka = {
  hosts: []
}
```

that way we will have more unified way to pass all the zk and kf settings

dwillett:
Ah okay, great. That all sounds good to me.

mensfeld:
@dwillett  for now please use the solution from the reported issue - I will update it and merge it in 2 weeks because now I'm heading to Japan (once I'm back I will have more time to update everything). Sorry for the inconvenience.

mensfeld:
@dwillett I'm back ;) will work on that soon

PR #77, opened 2016-04-18
haml-lint changed name to haml_lint

potomak:
Fixes on `bundle install`:

    An error occurred while installing haml-lint (0.13.0), and Bundler cannot continue.
    Make sure that `gem install haml-lint -v '0.13.0'` succeeds before bundling.

See also: https://github.com/polishgeeks/polishgeeks-dev-tools/issues/58.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,44 @@
+AllCops:
+  TargetRubyVersion: 2.3
+  Exclude:
+    - bin/**/*
+    - db/**/*
+    - .gemspec/**/*
+    - .bundle/**/*
+    - vendor/**/*
+    - config/**/*
+    - script/**/*
+    - !ruby/regexp /old_and_unused\.rb$/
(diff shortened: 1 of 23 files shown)


mensfeld:
@potomak  great work! could you just remove the .rubocop file? It is unnecessary since we ship with polishgeeks-dev-tools .rubocop file

potomak:
Well, I can't without breaking the build because of:

```
RSpec/AnyInstance:
  Enabled: false
RSpec/MultipleDescribes:
  Enabled: false
RSpec/InstanceVariable:
  Enabled: false
RSpec/FilePath:
  Enabled: false
```

Is there another way to partially override polishgeeks-dev-tools' rubocop configuration?

mensfeld:
@potomak that is true - we can use custom rubocop for now, since in few months we will drop devtools gem in favour of something new that is more independent. I will review this today

mensfeld:
Added to #78 branch

Issue #82, opened 2016-05-25
Karafka autodiscovery fails upon caching of configs


mensfeld:
Will be fixed with #81

Issue #81, opened 2016-05-24
Switch config management to dry configurable


mensfeld:
https://github.com/dry-rb/dry-configurable

mensfeld:
Rejected: does not support lazy loaded values on request

mensfeld:
Patched and ready - will also fix #82

PR #83, opened 2016-05-25
#81, #82

mensfeld:
@agwozdowski  @pavlo-vavruk  please review


lib/karafka/logger.rb
```diff
@@ -13,7 +13,7 @@ class Logger < ::Logger
 
     class << self
       # Builds a logger with appropriate settings, log level and environment
```
pavlo-vavruk: Maybe change docs to `Returns a logger instance with ..` once method does not have `build` name anymore?

mensfeld: done

spec/lib/karafka_spec.rb
```diff
@@ -3,82 +3,15 @@
...
-
-  describe '#logger=' do
-    let(:logger) { double }
-
-    it 'assigns logger' do
-      subject.logger = logger
-      expect(subject.instance_variable_get(:'@logger')).to eq logger
+  describe '.logger' do
+    it 'expect to use app logger' do
+      expect(subject.logger).to eq described_class::App.config.logger
```
pavlo-vavruk: @mensfeld I am not pretty sure, but maybe if we use `described_class` as a `subject` so we should use `subject::App.config.logger` and `subject::App.config.monitor` in the test below. What do  you think?

mensfeld: Not really - If by some reason it would start to be an instance (not a class), we would still refer internals based on the class.

pavlo-vavruk: Hence, everything looks great to me 👍

PR #85, opened 2016-06-02
Changed template to work with how the zk gem expects it

jwkoelewijn:
Original code resulted in:

karafka/lib/karafka/connection/broker_manager.rb:41:in `zk': undefined
  method `hosts' for ["127.0.0.1:2181"]:Array (NoMethodError)

when starting using 'bundle exec karafka server'


mensfeld:
Hehehe @agwozdowski fogot that ;) thank you!

PR #86, opened 2016-07-01
Added a zookeeper 'namespace' option.

jwkoelewijn:
Zookeeper instances can be namespaces, meaning that all znodes are
prefixed with the same name. This means that broker information would be
found at /namespace/brokers/ids instead of /brokers/ids.

This feature branch adds support for this, defaulting to an empty
namespace. Specs have been added to test the functionality.

PR #89, opened 2016-07-29
Jwkoelewijn zookeeper namespace support

mensfeld:
@agwozdowski @pavlo-vavruk  please review. This is based on work of @jwkoelewijn from #86 - thank you very much!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Karafka framework changelog
 
-## 0.4.1-head
+## 0.4.2-head
+- Added chroot option to Zookeeper options
+- Moved BROKERS_PATH into config from constant
+
+## 0.4.1
 - Explicit throw(:abort) required to halt before_enqueue (like in Rails 5)
 - #61 - Autodiscover Kafka brokers based on Zookeeper data
 - #63 - Graceful shutdown with current offset state during data processing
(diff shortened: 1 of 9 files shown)


agwozdowski:
👍

jwkoelewijn:
No problem at all :)

jwkoelewijn:
Very awesome this functionality was added, and now I finally had the time to actually try it on our staging environments. Turns out we forgot to change the way Poseidon::ConsumerGroup registers itself within Zookeeper.

The current solution does indeed find the Kafka Brokers within a chrooted Zookeeper, but when the consumer-groups are being registered, this chrooted environment is completely ignored, resulting in a quite confusing Zookeeper topology :)

I will try and see if I can supply a fix, but I see that poseidon_cluster is deprecated in favour of karafka, but karafka is still depending on it.... I might just create a fork of poseidon_cluster and make a pull-request here pointing to that or something like that... what do you reckon?

Btw, if I should've posted this under issues, please tell me and I will move it there.

JW

mensfeld:
@jwkoelewijn  Karafka is dependent on it until the kafka-ruby consumer is produciton ready. Unfortunately they still have more issues than Poseidon, that's why we stay with it. Meanwhile you can try to patch poseidon cluster here (as a part of Karafka).

Issue #87, opened 2016-07-28
Reconsume mode with crone for better Rails/Rack integration


mensfeld:

mensfeld:
@agwozdowski please close it once it is merged

PR #90, opened 2016-08-03
#87

mensfeld:
@agwozdowski  @pavlo-vavruk  please review and merge if ok :)


lib/karafka/server.rb
```diff
@@ -0,0 +1,46 @@
...
+      def bind_on_sigint
+        process.on_sigint do
+          Karafka::App.stop!
+          exit
+        end
+      end
+
+      # What should happen when we decide to quit with sigquit
+      def bind_on_sigquit
+        process.on_sigquit do
```
pavlo-vavruk: @mensfeld maybe, to get rid of duplication - move it to separate private method
```rb
def stop_app
  lambda do
    Karafka::App.stop!
    exit
  end
end
```
and use it 

```rb
def bind_on_sigint
  process.on_sigint(&stop_app)
end

def bind_on_sigquit
  process.on_sigquit(&stop_app)
end
```

Or is it redundant, how do you think?

mensfeld: it is redundant on purpose. If we would add more siqnal handlers than each of them would be separated

pavlo-vavruk:
👍

mensfeld:
thank you @agwozdowski  :)

Issue #76, opened 2016-04-10
Configs validator


mensfeld:
Now since everything is postloaded after user code, we could ad a config validator to check that allr equirements are met.

mensfeld:
Step 1 in #78

mensfeld:
@agwozdowski  we could use http://dry-rb.org/gems/dry-validation/ for that. Then we could define a schema that we could validate against and check that required things are present and that they have a proper format.

mensfeld:
This requires updating dry-configuration and removing / changing the autodiscovery on a config level (patched from old dry-config)

PR #91, opened 2016-08-04
Add chroot to zookeeper hosts when it is set.

jwkoelewijn:
The queue consumers pass the zookeeper host including the
chroot to the Poseidon::ConsumerGroup, which in turn uses
the ZK gem, which accepts chroots as part of the host url.

Extra care is taken to remove potential extra leading and
trailing slashes in the chroot. The zookeeper hosts array
is memoized, so the calculations will only be performed once

### DIFF
--- a/lib/karafka/connection/queue_consumer.rb
+++ b/lib/karafka/connection/queue_consumer.rb
@@ -69,7 +69,7 @@ module Karafka
         @target ||= Poseidon::ConsumerGroup.new(
           @route.group,
           ::Karafka::App.config.kafka.hosts,
-          ::Karafka::App.config.zookeeper.hosts,
+          zookeeper_hosts,
           @route.topic,
           socket_timeout_ms: (::Karafka::App.config.wait_timeout + TIMEOUT_OFFSET) * 1000,
           # How long should we wait for messages if nothing is there to process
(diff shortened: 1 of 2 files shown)


mensfeld:
@jwkoelewijn  thank you for your contribution. I will review and merge it once #87 it reviewed and merged (one thing at a time :D)

lib/karafka/connection/queue_consumer.rb
```diff
@@ -83,6 +83,24 @@ def target
...
+        end
+        @zookeeper_hosts
+      end
+
+      # @return zookeeper chroot with leading and trailing slashes removed or nil when
+      #   it is not set
+      def zookeeper_chroot
+        chroot = ::Karafka::App.config.zookeeper.chroot
+        return '' unless chroot
+        '/' + chroot.split('/').reject(&:empty?).join('/')
```
mensfeld: we will remove it, as there's a #76 that will ensure that we have a proper chroot provided :) (I will change - comment just as a reference)

mensfeld:
@jwkoelewijn  sorry for the delay. Will review this week.

mensfeld:
Hey - the new version does no longer that feature since the autodiscovery is based only on kafka - zookeeper is no longer required.

PR #102, opened 2016-10-03
fix error on startup from waterdrop

MikkelRBech:
doing a clean install then:
bundle exec karafka
gave:
NoMethodError: undefined method hosts' for {:hosts=>["127.0.0.1:9092"]}:Hash from waterdrop: lib/karafka/setup/configurators/water_drop.rb:12:inblock in setup'


mensfeld:
Totally forgot, that waterdrop is not yet using dry stuff. Thank you!

PR #109, opened 2016-10-14
README changes

Geesu:
Just fixed a worker that I believe should have been named something else. And a few text additions/changes to read better.


mensfeld:
Thank you! :-)

Issue #98, opened 2016-10-02
Use parser when responding on a topic


mensfeld:
By defualt should use JSON, but we should be able to assing a different on a responder level per topic

PR #116, opened 2016-10-31
#98

mensfeld:
@agwozdowski @pavlo-vavruk


agwozdowski:
👍  I'll add additional description about parser in example app issue
https://github.com/karafka/karafka-example-app/issues/18

PR #118, opened 2016-11-16
Added dot in topic validation format

Emile-Filteau:
__What are you trying to accomplish with this PR?__
Dots are valid in Kafka topic syntax. After starting a project with Karafka, i found out i couldn't map any route to my topics since they are namespaced using dots. 

__How is this accomplished?__
Adding the `\.` option in the `NAME_FORMAT` regex fixed my issue without causing any further problem.

I also tweaked the tests a bit by removing the `rand(...).to_s` to a valid string containing alphanumeric character, an underscore, a dash and a dot.

@mensfeld


mensfeld:
Hello @Emile-Filteau  - thank you for this contribution. I hade the dot in mind but since I don't use the dot format, it wasn't a priority for me. Thank you once again! :)

PR #123, opened 2016-12-15
update readme section about custom monitors which are now singletons

kevinhughes27:
I recently implemented a custom monitor following the example in the readme. Everything was super straight forward except it seems the base class was changed to a singleton after the docs were written. Running the code as is produces:

```
NoMethodError: private method `new' called for CustomMonitor:Class
```

I updated the readme to reflect how to construct the singleton object.

### DIFF
--- a/README.md
+++ b/README.md
@@ -665,7 +665,7 @@ class App < Karafka::App
   setup do |config|
     # Other setup stuff...
     config.logger = MyCustomLogger.new
-    config.monitor = CustomMonitor.new
+    config.monitor = CustomMonitor.instance
   end
 end
 ```


mensfeld:
Hello @kevinhughes27,

First of all, thank you for your contribution! :+1:  and of course for using Kafka. I hope you have as much fun using it as I had designing and building it :)

That was a great catch and I've even documented a reason why it became a singleton:

```
This class acts as a singleton because we are only permitted to have single monitor  per running process (just as logger)
```

Great catch!

kevinhughes27:
Thanks! It has been fun so far :)

Issue #96, opened 2016-09-29
Karafka server - possiblity to run it only for a certain consumer groups


mensfeld:
Similarly to how it is done in Sidekiq:

q, --queue QUEUE[,WEIGHT]       Queues to process with optional weights

mensfeld:
@mach-kernel  any idea on the DSL for that? Should we build up some sort of orchestration or just add an CLI option and leave the rest to the users (I would prefer the second one)?

mach-kernel:
We could add an `--only-topics=c,s,v` kind of flag for bare min functionality. I think that anything more would probably involve making DSL for it, but I don't really know how to represent that parity since I think each Karafka server process would then at minimum need to be aware of what topics its peers are subscribed to and that kind of sync seems to be an exercise we don't need to do at the current time. 

Agree?

mensfeld:
@mach-kernel that is good. I've figured out a way to do it easily I think and I should be able to make it for 0.6 :)

PR #124, opened 2016-12-16
Allow options into responder (fixes #95)

### LINKED ISSUE #95: Allow passing partition options into responder

thumbnail:
In our application we're quite dependant on partition support, so i added it :)

I chose to add support for waterdrop options per `respond_to`. That way the options can be based on the data. The proposed solution introduces a bit more complexity in the message buffer. A topic is no longer linked to just the message, but an array/pair of the data and the waterdrop options.

Let me know what you think

### DIFF
--- a/lib/karafka/base_responder.rb
+++ b/lib/karafka/base_responder.rb
@@ -16,6 +16,15 @@ module Karafka
   #     end
   #   end
   #
+  # @example Responding to a topic with extra options
+  #   class Responder < BaseResponder
+  #     topic :new_action
+  #
+  #     def respond(data)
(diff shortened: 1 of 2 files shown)


mensfeld:
Hey @thumbnail that is really, really nice! We will review it shortly, check compatibility with waterdrop, because I'm not sure whether or not is supports passing options to the driver and merge it once we're done.

Great job! :)

mensfeld:
@thumbnail everything is awesome. No waterdrop changes required. I will update readme a bit to match new options so people will be aware.

Btw since you use partitions then probably you will be happy about #96 - I will implement that soon

Issue #128, opened 2017-02-02
Monitor caller_label not working with super on inheritance


kwestein:
👋 

The base Monitor class defines the [caller_label method](https://github.com/karafka/karafka/blob/master/lib/karafka/monitor.rb#L73) as
````
index = self.class.ancestors.index(Karafka::Monitor) + 1
caller_locations(index, 2)[1].label
````

I am wondering why the choice was made to have the method return ````caller_locations(index, 2)[1].label````? As an example, when monitor.notice is called from within a block, these values are returned:
````
caller_locations(index, 2)[0].label = 'notice'
caller_locations(index, 2)[1].label = `block in notice_signal'
````

In the ReadMe, the [example monitor with new relic support](https://github.com/karafka/karafka#example-monitor-with-newrelic-support) calls method by name based on the string value returned by caller_label. Given that the defined method would be called  ````notice```` not ever ````block in notice signal````, should this method not be returning caller_locations(index, 2)[0].label? I overrode it to do so in the app monitor we are using in our app and it works perfectly.

Let me know if I am missing something and it behaves this way intentionally. Otherwise I am happy to make the change. @mensfeld 

cc: @kevinhughes27

mensfeld:
Hello @kwestein  :-) thank you for your interest in Karafka. This is a really interesting question. I will investigate that and get back to you max on Monday it that's not an asap

mensfeld:
Btw you can always ask questions and get support faster on our gitter account: https://gitter.im/karafka/karafka

mensfeld:
@kwestein  @kevinhughes27  could you please provide me with your monitor? To investigate it further I will have to have your monitor on which I could play. Caller label gets tricky when it comes down to the inheritance so. having yours and seeing how you use it would be a big help for me.

kwestein:
This is what we currently have:

````
# frozen_string_literal: true
module Karafka
  class AppMonitor < Karafka::Monitor
    def notice(caller_class, options = {})
      super
      return unless respond_to?(caller_label, true)
      send(caller_label, options)
    end

    def notice_error(caller_class, e)
      super
    end

    private

    def caller_label
      index = self.class.ancestors.index(Karafka::Monitor) + 1
      caller_locations(index, 2)[0].label
    end

    def consume(params)
      content = JSON.parse(params[:content])
      logger.info("#{params[:controller]} received kafka event: #{content}")
      record_count params[:topic], content.try(:[], "event_name"), __method__
    end

    def perform_async(params)
[...]

PR #129, opened 2017-02-06
#128

mensfeld:
Solution to #128 please read the thread, review and comment :)

### DIFF
--- a/lib/karafka/monitor.rb
+++ b/lib/karafka/monitor.rb
@@ -73,8 +73,19 @@ module Karafka
     def caller_label
       # We need to calculate ancestors because if someone inherits
       # from this  class, caller chains is longer
-      index = self.class.ancestors.index(Karafka::Monitor) + 1
-      caller_locations(index, 2)[1].label
+      index = self.class.ancestors.index(Karafka::Monitor)
+      # caller_locations has a differs in result whether it is a subclass of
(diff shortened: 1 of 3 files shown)


lib/karafka/monitor.rb
```diff
@@ -73,8 +73,19 @@ def caller_exceptions_map
...
+      index = self.class.ancestors.index(Karafka::Monitor)
+      # caller_locations has a differs in result whether it is a subclass of
+      # Karafka::Monitor, the basic Karafka::Monitor itself or a super for a subclass.
+      # So to cover all the cases we need to differentiate.
+      # @see https://github.com/karafka/karafka/issues/128
+      # @note It won't work if the monitor caller_label caller class is defined using
+      #   define method
+      super_execution = caller_locations(1, 2)[0].label == caller_locations(1, 2)[1].label
+
+      scope = super_execution ? 1 : nil
```
pavlo-vavruk: We can write it in one line without this `||=`
Am not sure whether it is more readable. For me - it is, because we can easily read when it has to be equal to 1.
`scope = super_execution || !index.positive? ? 1 : 0 `

mensfeld: @pavlo-vavruk  rubocop will compain ;)

PR #141, opened 2017-04-23
Add topic prefix

mach-kernel:
Other half of [waterdrop #25](https://github.com/karafka/waterdrop/pull/25)

### DIFF
--- a/lib/karafka/routing/route.rb
+++ b/lib/karafka/routing/route.rb
@@ -47,9 +47,11 @@ module Karafka
         (@group ||= "#{Karafka::App.config.name.underscore}_#{topic}").to_s
       end
 
-      # @return [String] route topic - this is the core esence of Kafka
-      def topic
-        @topic.to_s
+      # @return [String] route topic - this is the core essence of Kafka
+      # @note If kafka.topic_mapper is set, the topic will be set to the return
+      #       value of that block
+      def topic(mapper = ::Karafka::App.config.kafka.topic_mapper)
+        (mapper.is_a?(Proc) ? mapper.call(@topic) : @topic).to_s
       end
 
       # @return [Class] Class (not an instance) of a worker that should be used to schedule the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 0.5.0.2-head
 - #132 - When Kafka is gone, should reconnect after a time period
 - #136 - new ruby-kafka version + other gem dumps
+- #141 - Adds `topic_prefix` config flag
 
 ## 0.5.0.2
 - Gems update x3
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,9 @@ source 'https://rubygems.org'
 
 gemspec
 
+# TODO: move back to *.gemspec after merge
+gem 'waterdrop', git: 'https://github.com/mach-kernel/waterdrop.git', branch: 'add-topic-prefix'
+  
 group :development, :test do
   gem 'polishgeeks-dev-tools'
   gem 'timecop'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/mach-kernel/waterdrop.git
-  revision: b576a4df99f769be67af3b724c6c691c53a3931c
+  revision: 54fb908f11ae8d8a727f4f0ec513469a0dbcc99e
   branch: add-topic-prefix
   specs:
     waterdrop (0.3.2.4)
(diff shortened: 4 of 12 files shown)


mensfeld:
@mach-kernel it's me again. :) Thank you for this PR - I have 2 doubts about that PR:

### Topic prefix effective range
the way it is implemented makes it only partially functional as it adds the prefix to the topic (which is awesome), but then does not hide the original topic name details inside of Karafka. I mean you define the prefix but you still have to remember about it in responders and you have the full topic name inside the params[:topic] value. IMHO (dunno what you guys think) it would have more sense if it would be a complex prefix usage that takes care of all of the cases - I mean let's assume we decide to use CloudKarafka (they stole our name btw :( ) and we namespace: we should operate everywhere in the Karafka framework scope with "norma" non-namespaced names and the framework should take care of the rest. If we would implement this that way, then switching to a different provider that does not require prefixes would not require any extra work.

### Topic prefix functionality
I think that it would be better to support more elastic way of doing things. Something like that would be more elastic:
```ruby
config.topic_mapper = -> (original_name){ "my_prefix_#{original_name}" }
```
and could support topic name transformation, per topic options etc.

@agwozdowski  @pavlo-vavruk please feel free to leave your remarks as well

mensfeld:
@mach-kernel I can easily implement the above things based on your code if you're too busy or just don't know where to start :)

mach-kernel:
@mensfeld I love the topic mapper idea! I'll get started on it and will report back in a few days. Will totally loop you in with questions. Thanks for considering the PR and this new feature! 😸 

Sorry to hear that they stole your name 😞

mensfeld:
@mach-kernel  feel free to ask as many questions as you want. You cn find me on gitter quite often as well. This is going to be big, since it needs a new topic mapper concept that will work on many layers so no rush :-)

lib/karafka/routing/route.rb
```diff
@@ -47,9 +47,11 @@ def group
...
       end
 
-      # @return [String] route topic - this is the core esence of Kafka
-      def topic
-        @topic.to_s
+      # @return [String] route topic - this is the core essence of Kafka
+      # @note If kafka.topic_mapper is set, the topic will be set to the return
+      #       value of that block
+      def topic(mapper = ::Karafka::App.config.kafka.topic_mapper)
+        (mapper.is_a?(Proc) ? mapper.call(@topic) : @topic).to_s
```
mensfeld: I think it would be better to check if it can be called not if it is a proc - that way I could inject modules or instances of classes to do more complex topic mappings.

mach-kernel: Would it be enough to expect the user to conform to an API with a `:call` or `:map` method?

mensfeld:
I think we should drop it in waterdrop totally. If we would leave waterdrop as an internal karafka detail and move to do it only in karafkas routing engine + responders , waterdrop would always receive proper mapped topics. Karafkas default and recommended way of sending messages is with responders. If you give me 24h I will create a description of how it could work. It would be super elastic and extendable. Sorry for not figuring that earlier.

mach-kernel:
@mensfeld I would actually need the features on both ends; I have a REST API built with Grape that publishes messages to Kafka. They are consumed by `karafka` which then runs each message through whatever internal pipelines we have made. I could wrap `waterdrop` in some sort of adapter but so many Kafka SaaS hosts do this that it might be useful to more than just me, so don't base your decision on me and do what's best for the framework -- but I thought I would illustrate my use case. :smile_cat: 

I'm totally down to hear your plan :+1:

mensfeld:
@mach-kernel can't you use responders as a an API client? This would make things so much better for you as it controls the flow much better. Let me give you an example of how it could work:

## Configuring topic mapper on Karafka
```ruby
class App < Karafka::App
  setup do |config|
    config.topic_mapper = ->(topic) { "your_name.#{topic}" }
  end
end
```

and this would be all. You could normally describe routes without prefix but also use responders like so:

```ruby
# The json parser will be default soon and there will be the .call alias on a class
CreatedResponder.new(Karafka::Parsers::Json).call(model)
```

and the responder could look that way:
```ruby
  module Repositories
    class CreatedResponder < KarafkaResponder
      topic :repositories_created

      def respond(repository)
        respond_to :repositories_created, repository
      end
    end
  end
end
```

Notice, that I don't use the user namespace on topics directly - it could be easily remapped both ways (for incoming and outgoing messages) if we would leave it in the router + responders part.

Waterdrop is used under the responders engine and if we would leave it "pure" we could make that happen without any additional logic.

mach-kernel:
@mensfeld It hadn't occurred to me to use the responders directly. This pattern sounds great -- and I'm not too far along in the development of what I'm working on to cut `waterdrop` out for `karafka`. 

Would you like me to implement this pattern in this PR? Seems pretty clear!

mensfeld:
@mach-kernel yeah - people tend to forget that there's this great responders api that provides an runtime validation layer that makes things a bit easier :) 

 I would be more than happy if you would be willing to implement the topic mapper logic inside of Karafka. I will close this issue and create a new one (will update it in 1-2 days) and ping you once it is ready :)

Issue #138, opened 2017-04-16
You should be able to have no worker class defined if inline_mode is being used everywhere


kwestein:
When using `inline_mode = true` everywhere, you should not need to define a worker since everything would be performing inline. Yet I am still seeing `Karafka::Errors::BaseWorkerDescentantMissing` when I try to remove our worker definition. 

@mensfeld cc: @kevinhughes27 @SebastianSzturo

mensfeld:
- Should not validate base worker descendants at all
- Should not try to build anything at all

PR #144, opened 2017-04-28
Resolving issues #138 #76

agwozdowski:


mensfeld:
Hey @agwozdowski  thank you :-) great work! I will be adding comments during current Holiday, so they will occur from time to time.

README.md
```diff
@@ -112,13 +112,13 @@ Karafka has following configuration options:
 | Option                        | Required | Value type        | Description                                                                                                |
 |-------------------------------|----------|-------------------|------------------------------------------------------------------------------------------------------------|
 | name                          | true     | String            | Application name                                                                                           |
-| redis                         | true     | Hash              | Hash with Redis configuration options                                                                      |
+| redis                         | false    | Hash              | Hash with Redis configuration options. It is required if inline_mode is true.                              |
```
mensfeld: inlne_mode is off I guess ;) if it is on, we don't have to have redis settings at all

lib/karafka/app.rb
```diff
@@ -12,7 +12,10 @@ def setup(&block)
       # Sets up all the internal components and bootstrap whole app
       # We need to know details about routes in order to setup components,
       # that's why we don't setup them after std setup is done
+      # @raise [Karafka::Errors::WrongConfiguration] raised when configuration
```
mensfeld: Karafka::Errors::InvalidConfiguration

lib/karafka/errors.rb
```diff
@@ -52,5 +52,8 @@ class TopicMultipleUsage < BaseError; end
 
     # Raised when we didn't use a topic that was defined as non-optional (required)
     class UnusedResponderRequiredTopic < BaseError; end
+
+    # Raised when configuration doesn't match with validation schema
+    class WrongConfiguration < BaseError; end
```
mensfeld: Please change to Karafka::Errors::InvalidConfiguration - we use for other errors the "Invalid" prefix so it would be good for consistency to keep it that way.

lib/karafka/routing/route.rb
```diff
@@ -56,6 +56,7 @@ def topic
       #   background job
       # @note If not provided - will be built based on the provided controller
       def worker
+        return nil if inline_mode
         @worker ||= Karafka::Workers::Builder.new(controller).build
```
mensfeld:
```
@worker ||= inline_mode ? nil : Karafka::Workers::Builder.new(controller).build
```

lib/karafka/setup/config.rb
```diff
@@ -37,7 +37,7 @@ class Config
         # Connection pool size for producers. Note that we take a bigger number because there
         # are cases when we might have more sidekiq threads than Karafka routes (small app)
         # or the opposite for bigger systems
-        setting :size, -> { [::Karafka::App.routes.count, Sidekiq.options[:concurrency]].max }
+        setting :size, [::Karafka::App.routes.count, Sidekiq.options[:concurrency]].max
```
mensfeld: Are you sure we can do this that way? I mean do that without a block?

lib/karafka/setup/config.rb
```diff
@@ -95,6 +95,15 @@ def setup_components
...
           end
         end
+
+        # Validate config based on ConfigurationSchema
+        # @raise [Karafka::Errors::WrongConfiguration] raised when configuration
+        #   doesn't match with ConfigurationSchema
+        def validate!
+          validation_result = Karafka::Setup::ValidationSchema.call(config.to_h)
+
+          raise Errors::WrongConfiguration, validation_result.errors if validation_result.failure?
```
mensfeld: I would prefer
return true if validation_result.success?

raise...

lib/karafka/setup/validation_schema.rb
```diff
@@ -0,0 +1,43 @@
+module Karafka
+  module Setup
+    # Schema with validation rules for all configuration
+    ValidationSchema = Dry::Validation.Schema do
```
mensfeld: All dry validation schemas are validation schemas. I would suggest calling it Karafka::Setup::ConfigSchema (or ConfigurationSchema depending on the existing convention)

lib/karafka/setup/config.rb
agwozdowski: We can, I think. It's default value for size setting. Added validator about size which should be an integer.

agwozdowski:
Thanks, @mensfeld! Done :)

lib/karafka/app.rb
mensfeld: Still not updated :(

lib/karafka/setup/config.rb
mensfeld: I disagree and I know why - when config is being build dynamically, it is set based on the routes and we know nothing about them in this moment. That's why it is being delegated to be decided later on. This needs to stay as a proc

PR #156, opened 2017-06-24
Handle SIGTERM as a shutdown command for kafka server to support Heroku deployment

webandtech:
We use Karafka with Heroku Kafka and noticed that Kafka Server does not properly shut down on Heroku because SIGTERM is used instead of SIGQUIT or SIGINT. This PR simply adds support for SIGTERM as well.

### DIFF
--- a/lib/karafka/process.rb
+++ b/lib/karafka/process.rb
@@ -6,7 +6,7 @@ module Karafka
 
     # Signal types that we handle
     HANDLED_SIGNALS = %i(
-      SIGINT SIGQUIT
+      SIGINT SIGQUIT SIGTERM
     ).freeze
 
     HANDLED_SIGNALS.each do |signal|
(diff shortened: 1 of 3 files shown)


mensfeld:
Hello @webandtech! I didn't even realize that Karafka works with Heroku! Great news! I will update readme with it :) Meanwhile thank you for your contribution. I will review it next week but initially looks great!

webandtech:
Great! And thanks for a great framework. It works great with Heroku Kafka, for your readme, with Heroku Kafka you have to configure Karafka like this:
```
config.kafka.hosts = ENV['KAFKA_URL'].split(',') # Convert CSV list of broker urls to an array
config.kafka.ssl.ca_cert = ENV['KAFKA_TRUSTED_CERT']
config.kafka.ssl.client_cert = ENV['KAFKA_CLIENT_CERT']
config.kafka.ssl.client_cert_key = ENV['KAFKA_CLIENT_CERT_KEY']
```

mensfeld:
Thank you once again. Great job!

PR #157, opened 2017-06-24
Allow karafka root to be specified from ENV

webandtech:
Allow using `ENV['KARAFKA_ROOT_DIR']` to specify the root directory for Karafka. We use this so we can have all of our Karafka classes and logic in a subdirectory of a parent Rails app, but still share the same Gemfile. The Rails app can then be loaded without Karafka's classes and dependencies.


mensfeld:
@webandtech great job! Thank you and sorry for the merging delay.

webandtech:
@mensfeld any plans to push an updated gem to rubygems?

mensfeld:
@webandtech tomorrow I will update the README and I can push 0.5.0.3 especially for you <3
I can push it even today if you update the README on Heroku section #158

Issue #186, opened 2017-08-10
Support ruby-kafka `ssl_ca_cert_file_path` config


webandtech:
I was looking at the 0.6.0 changes and I noticed that the `ssl_ca_cert_file_path` option from ruby-kafka isn't included in the updated configs. Could we add that? It is now required to use karafka with Heroku Kafka. See: https://devcenter.heroku.com/articles/ca-cert-rotation-kafka#ruby-kafka

Simply allowing `config.kafka.ssl_ca_cert_file_path` to be set would allow the proper configuration.

mensfeld:
@webandtech thank you for your contribution! I will review the changes max in 2 days (probably will do it faster)

webandtech:
@mensfeld it will have to be changed for 0.6.0. When are you planning to release it?

mensfeld:
@webandtech that is a tricky question. 0.6 has been running on production for me for a while so I consider it stable BUT the docs aren't. I need to update them and due to the number of changes, I need to go thought all of them :( So I expect to make a rubygems release of 0.6 in 2-3 weeks.

webandtech:
@mensfeld no problem. You could consider publishing RC versions of this and the updated waterdrop gem. See: http://guides.rubygems.org/patterns/#prerelease-gems i.e you could release `0.6.0.rc1`

mensfeld:
@webandtech  will do. Waterdrop is already bumped and released as 0.4 > https://rubygems.org/gems/waterdrop

mensfeld:
@webandtech  will do an RC after the weekend.

PR #187, opened 2017-08-10
add support for ruby-kafka `ssl_ca_cert_file_path` config

webandtech:
See #186 

The ability to load the ca cert from a file is required to support Heroku Kafka due to a change they are rolling out where multiple ca certs may be present on the brokers. Also see: https://devcenter.heroku.com/articles/ca-cert-rotation-kafka


mensfeld:
@webandtech  thank you! I'm personally not using Heroku, that's why we might be a bit behind their changes. Really, really appreciate your work!

mensfeld:
@webandtech one more thing. Would you be willing to look @ the heroku deployment section to  check, whether or not it is up 2 date? https://github.com/karafka/karafka/wiki/Deployment

Issue #196, opened 2017-08-18
Topic mapper does not map topics when subscribing


webandtech:
We are attempting to use a topic mapper to allow us to transparently add a prefix to the topics without having to explicitly add it when drawing routes or configuring responders, however we found that even with a topic mapper defined in setup, Karafka would still subscribe to the bare topics (unprefixed).

In this situation:

```
# topic_mapper.rb
module StagingTopicMapper
    PREFIX = 'staging_'

    def self.incoming(topic)
      topic.to_s.gsub(PREFIX, '')
    end

    def self.outgoing(topic)
      "#{PREFIX}#{topic}"
    end
  end
```

```
  # app.rb
  consumer_groups.draw do
      topic :my_topic
    end
```

Karafka should subscribe to the topic `staging_my_topic` but it instead subscribes to `my_topic` as if there was no topic mapper registered.

PR to follow!

mensfeld:
Works on my side. Feel free to reopen if you have any more problems with it. I will also add an issue to create more specs to cover that. That way, we won't lose this case never again.

PR #197, opened 2017-08-18
fix [#196] incorrect subscriptions are created when topic mapper is used

webandtech:
fixes [#196]

In the below situation, karafka will now correctly subscribe to `staging_my_topic`. Also fixes an issue where when a topic mapper is in place the `@topic` instance variable on the message would get replaced, leading to the message to be processed over and over as it would never be committed as read.

```
# topic_mapper.rb
module StagingTopicMapper
    PREFIX = 'staging_'

    def self.incoming(topic)
      topic.to_s.gsub(PREFIX, '')
    end

    def self.outgoing(topic)
      "#{PREFIX}#{topic}"
    end
  end
```

```
  # app.rb
  consumer_groups.draw do
      topic :my_topic
    end
```

### DIFF
--- a/lib/karafka/connection/config_adapter.rb
+++ b/lib/karafka/connection/config_adapter.rb
@@ -80,7 +80,7 @@ module Karafka
             settings[setting_name] = setting_value
           end
 
-          [topic.name, sanitize(settings)]
+          [Karafka::App.config.topic_mapper.outgoing(topic.name), sanitize(settings)]
         end
 
         private
--- a/lib/karafka/connection/messages_processor.rb
+++ b/lib/karafka/connection/messages_processor.rb
@@ -24,7 +24,7 @@ module Karafka
           controller = Karafka::Routing::Router.build("#{group_id}_#{mapped_topic}")
           handler = controller.topic.batch_processing ? :process_batch : :process_each
 
-          send(handler, controller, mapped_topic, kafka_messages)
+          send(handler, controller, kafka_messages)
           # This is on purpose - see the notes for this method
           # rubocop:disable RescueException
         rescue Exception => e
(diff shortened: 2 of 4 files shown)


webandtech:
@mensfeld If this fix isn't needed and the issue is I am using the topic mapper incorrectly please let me know!

mensfeld:
Hello @webandtech, first of all, thank you for your contribution and taking your time to play with 0.6.0.rc1. I will review this shortly and get back to you :-)

lib/karafka/connection/config_adapter.rb
```diff
@@ -80,7 +80,7 @@ def subscription(topic)
             settings[setting_name] = setting_value
           end
 
-          [topic.name, sanitize(settings)]
+          [Karafka::App.config.topic_mapper.outgoing(topic.name), sanitize(settings)]
```
mensfeld: great job on catching that one!

lib/karafka/connection/messages_processor.rb
```diff
@@ -38,34 +38,22 @@ def process(group_id, kafka_messages)
...
         # @param mapped_topic [String] mapped topic name
         # @param kafka_messages [Array<Kafka::FetchedMessage>] raw messages from kafka
-        def process_batch(controller, mapped_topic, kafka_messages)
-          messages_batch = kafka_messages.map do |kafka_message|
-            # Since we support topic mapping (for Kafka providers that require namespaces)
-            # we have to overwrite topic with our mapped topic version
-            # @note For the default mapper, it will be the same as topic
-            # @note We have to use instance_variable_set, as the Kafka::FetchedMessage does not
-            #   provide attribute writers
-            kafka_message.instance_variable_set(:'@topic', mapped_topic)
```
mensfeld: this remapping needs to happen, otherwise if you remap with staging_ you will get messages with staging_ back, which means that someone might have a business logic relying on the namespaced topic name instead of the proper unmapped to the env one (in your case). I will fix that once it is merged so no worries :)

mensfeld:
Hey @webandtech, thank you again. Looks great. I will merge it and make a small change to revert the remapping upon receiving. This is an amazing fix <3 Thank you!

lib/karafka/connection/messages_processor.rb
webandtech: The remapping was causing the message to be consumed in an endless loop in my testing as ruby-kafka was seeing the incorrect (non-actual) topic name. I think the message itself should always have the real topic name in my opinion and the karafka code interpreting it with the mapper as needed... perhaps I misunderstand why the message must be altered at this level?

mensfeld: @webandtech I know. I fixed that already on the param level :)

mensfeld: @webandtech already fixed in master. Would appreciate if you could verify that it is working.

webandtech: @mensfeld 👍 works for me now!

PR #199, opened 2017-08-21
#96

mensfeld:
So, since we now operate on consumer groups, this really allows us to listen on selected consumer groups on a given process not per topic but the idea is exactly the same.

Why on consumer group level? First of all I don't see a reason to support a really weird case, when we subscribe on some of the topics of a given consumer group, second of all, there always a per topic consumer group option, so if someone needs crazy thing, still can achieve.


mensfeld:
I even got coditsu a bit lower :D https://app.coditsu.io/karafka/repositories/karafka/builds/commit_builds

PR #202, opened 2017-08-23
Autoload

mensfeld:


mensfeld:
Have to merge it as I need it :rocket:

lib/initializer.rb
```diff
@@ -0,0 +1,58 @@
...
+
+Karafka::Loader.load!(Karafka.core_root)
+
+if rails_preloaded
+  # Railtie used to load Karafka stuff into Rails, so all routes, etc are available
+  # from Rails console and in other Rails related places
+  # @note This will load Karafka to Rails, not the other way around
+  Class.new(Rails::Railtie) do
+    config.to_prepare do
+      p 'aaaaaaaaaa'
```
thumbnail: Especially like this contribution to my rails app

mensfeld: @thumbnail it was fixed in less than 10 minutes. This whole change has been reversed and rc2 does not contain that. Really sorry for that. Autoload was a bad idea.

thumbnail: I noticed when deployment used a new version (not yet locked at a specific hash) with this change. Hadn't noticed the fix yet.

Whilst we're on the topic, We were using it in a non-standard setup (and not in a rails environment). And in order to extend autoload behavior, I had to monkey patch the loader (or write my own). might want to take that into consideration when designing a solution for this problem :).

mensfeld: @thumbnail first of all, thank you for using the github version. Probably shouldn't say that, but it's not the safest idea :D (especially when there's a rc2 already). Also there is no problem if you don't acknowledge it :D the autoloader rails feature had multiple corner cases,especially when combined with bare karafka and other frameworks. It is much easier to instruct users to uncomment 4 lines in karafka.rb file and add 1 line in their environment.rb file than to try to cover all the weird cases that emerge (especially, when people disable gem auto require in bundler, etc).

Once again: really sorry. I will take some steps to prevent that from happening again.

Issue #204, opened 2017-08-24
Long running controllers


mensfeld:
There might be cases in which someone wants to commit the offset manually, for example when aggregating data from multiple batches.

It would be good to provide a long running controllers, that would  be alive with an accumulator, etc) up until the buffer is full, then we could commit the offset do our final logic and nullify the controller with its buffers, etc.

mensfeld:
I need this, I will roll that out for 0.6 final.

Issue #147, opened 2017-05-09
Deattach Karafka from Sidekiq by extracting Sidekiq backend


mensfeld:
A lot of people use Karafka without Sidekiq in an inline mode. We could remove sidekiq as a dependency and allow users to use (or even implement) any async adapter as long as they should follow a simple API. We would create one for Sidekiq ofc. That way people could use Sidekiq Pro easier or Resque or anything they want.

mach-kernel:
I can take this @mensfeld 👍 

Sidekiq, Resque, and DelayedJob seem to be popular for this kind of thing. Can you think of any others?

mensfeld:
@mach-kernel feel free. However I think that this one requires some planning. I will try to describe my initial idea and some things we need to consider in 2 days. Then we can process this further. Are you ok with that?

mach-kernel:
@mensfeld sure thing!

mensfeld:
Note: it is still a WIP:

1. We need to consider an inline mode just an adapter as well
2. We need to consider, the serialization (interchanger problem)
3. New engine needs to support current batch mode and future batch mode (mass batch mode for batch processing)
4. What about workers? Now they are tightly couples with Sidekiq
5. Can we differ adapters in between topics allowing to use different (one per topic?). I think so, as then we could have both async adapters and sync adapter and replace the inline mode with it.
6. What about the CLI - the worker job won't longer work. Should we drop that totally or should we implement worker per adapter? (I would be in favour of dropping - everyone can run sidekiq ;) ) . If we drop, we need to update readme.

mensfeld:
@mach-kernel I think that we could do this step by step:
We could:
- remove worker-glass from dependencies (it's a really simple logic) and just maintain it as an additional feature for sidekiq adapter (adapter specific tool)
- remove karafka worker at all (this is also sidekiq specific) - it is not hooked up to the karafka-capistrano and should be fairly easy to migrate (as it is just alias for bundle exec sidekiq whatever)

mach-kernel:
From `worker-glass`:

- #perform (do work)
- #after_failure (ensure clean state)

I think it might not be a bad idea to make a base class that does a `#before_perform` step with an array of actions. It'd be a bit cleaner than chaining super (e.g. multiple prepend) calls and maybe ordering them could be helpful? At the very least it would match the `before_enqueue` pattern found in controllers. Standardizing this would be a good first step I think, to do along with removing `worker-glass`.

[...]

PR #206, opened 2017-08-25
#204 - Long running controllers

mensfeld:
Now we build a controller instance per batch, this disallows us from doing some cross-batch stuff like aggregating up until we have what we need to flush to DB and so on.

This PR aims to fix that and provide a single controller instance per topic partition. It also lowers the GC usage and makes things a bit faster.

Also, the topic ownership over controller is moved to a class level.

This PR contains also refactor that extracts topic configuration controller specific behaviors into modules that are included on demand only when needed.


mensfeld:
Added the #207 to this PR as well.

lib/karafka/base_controller.rb
```diff
@@ -95,33 +98,12 @@ def params_batch=(messages)
...
-        else
-          raise Errors::InvalidProcessingBackend, topic.processing_backend
-        end
-      end
-    end
-
-    # @note We want to leave the #perform method as a public API, but just in case we will do some
-    #   pre or post processing we use call method instead of directly executing #perform
+    # will call process method of a proper backend
     def call
```
mach-kernel: Much cleaner interface here

mensfeld: You know, that I did this mostly for your #147 ;) to simplify things (no rush with that one though)

lib/karafka/base_worker.rb
```diff
@@ -10,17 +10,18 @@ class BaseWorker
...
     # @return [Karafka::Controller] descendant of Karafka::BaseController that matches the topic
     #   with params_batch assigned already (controller is ready to use)
     def controller(topic_id, params_batch)
-      @controller ||= Karafka::Routing::Router.build(topic_id).tap do |ctrl|
-        ctrl.params_batch = ctrl.topic.interchanger.parse(params_batch)
-      end
+      topic = Karafka::Routing::Router.find(topic_id)
+      controller = topic.controller.new
+      controller.params_batch = topic.interchanger.parse(params_batch)
+      controller
```
mach-kernel: You don't have to make this change, but I always cannot resist some playful code golf :joy::

```ruby
topic.controller.new.tap { |c| c.params_batch = topic.interchanger.parse(params_batch) }
```

lib/karafka/controllers/sidekiq_backend.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Karafka
+  module Controllers
+    # Sidekiq backend that schedules stuff to Sidekiq worker for delayed execution
+    module SidekiqBackend
```
mach-kernel: 👍

lib/karafka/controllers/includer.rb
```diff
@@ -0,0 +1,22 @@
...
+  # Additional functionalities for controllers
+  module Controllers
+    # Module used to inject functionalities into a given controller class, based on the controller
+    # topic and its settings
+    # We don't need all the behaviors in all the cases, so it is totally not worth having
+    # everything in all the cases all the time
+    module Includer
+      # @param controller_class [Class] controller class, that will get some functionalities
+      #   based on the topic under which it operates
+      def self.call(controller_class)
```
mach-kernel: This can also be a great way for other libraries to build prefab stuff into controllers.

mensfeld: Yeah, we can come up with something more flexible in the future, but I agree, that this is a good start

Issue #210, opened 2017-09-15
LoadError: cannot load such file -- [...]/karafka.rb


filiptepper:
I'm following clean installation tutorial from the wiki. https://github.com/karafka/karafka/wiki/Getting-started

After `bundle install` I run `bundle exec karafka install` with the following error:

```bash
$ bundle exec karafka install
bundler: failed to load command: karafka ([...]/vendor/bundle/ruby/2.4.0/bin/karafka)
LoadError: cannot load such file -- [...]/karafka.rb
  [...]/vendor/bundle/ruby/2.4.0/gems/karafka-1.0.0/bin/karafka:4:in `require'
  [...]/vendor/bundle/ruby/2.4.0/gems/karafka-1.0.0/bin/karafka:4:in `<top (required)>'
  [...]/vendor/bundle/ruby/2.4.0/bin/karafka:23:in `load'
  [...]/vendor/bundle/ruby/2.4.0/bin/karafka:23:in `<top (required)>'
```

It appears that `karafka.rb` is necessary for the command to run (empty file worked fine).

Would you consider:
* adding a note in the tutorial about `karafka.rb` being required?
* or even better, adding a generator for default `karafka.rb` file when `karafka install` is executed?

mensfeld:
Yes and yes. We will fix that today. Thank you.

mensfeld:
Thank you once again for reporting this error. Fixed on master.

PR #211, opened 2017-09-15
#210

mensfeld:


mensfeld:
@filiptepper  feel free to review.

AS IS:
as described here: https://github.com/karafka/karafka/issues/210

TO BE:

in case of ```-h``` and ```install``` the ```karafka.rb``` is not required. For all other commands, if not present, ```Karafka::Errors::BootFileMissing``` exception is raised.

filiptepper:
❤️

PR #217, opened 2017-09-20
check RUBY_ENGINE constant if RUBY_VERSION is missing

filiptepper:
`spec_helper.rb` checks for `RUBY_VERSION` environment variable.

Perhaps it could fall back to `RUBY_VERSION` constant for `jruby?` check?


mensfeld:
Thank you.

PR #218, opened 2017-09-20
add configuration setting to control Celluloid's shutdown timeout

filiptepper:
At times our consumers take longer than default 30 seconds to process a batch of messages. We'd like to extend the default shutdown period (just like we have extended termination period for docker containers running these consumers).

Not sure if that should to be scoped by `celluloid`, but here goes.


mensfeld:
Funny thing - I was thinking about moving that to config 2 days ago. Thank you. I will make corresponding config rules today.

PR #220, opened 2017-09-21
t#219

mensfeld:
@filiptepper feel free to review :)


filiptepper:
Thanks! Just tested it on my code with custom mapper and it works as expected. 👌

Issue #225, opened 2017-10-14
Unable to override WaterDrop settings


ankitagupta12:
`Karafka::Setup::Configurators::WaterDrop` hardcodes the value for `send_messages` and `raise_on_failure` to true. This makes it difficult to override the settings inside an app to `false` for development environment.

```
::WaterDrop.setup do |water_config|
            water_config.send_messages = true
            water_config.raise_on_failure = true
            water_config.connection_pool = config.connection_pool
            .
            .
end
```

The values should ideally we configurable similar to `connection_pool` settings so that they can be defined in the config and overridden as necessary

ankitagupta12:
@mensfeld here's a PR for this issue: https://github.com/karafka/karafka/pull/226. let me know if this looks okay

mensfeld:
@ankitagupta12 as a temp workaround you could just overwrite it with your own settings after the ```App.boot!```

mensfeld:
@ankitagupta12 this is how it will work in 1.1:
```ruby
class App < Karafka::App
  setup do |config|
    config.kafka.seed_brokers = ::Settings.kafka.seed_brokers
    config.kafka.offset_commit_threshold = 30
    config.client_id = ::Settings.name
    config.monitor = KarafkaMonitor.instance
    config.backend = :inline
  end

  consumer_groups.draw do
    consumer_group 'a' do

      topic :a do
        backend :inline
        controller CountersController
        batch_processing true
        persistent true
        interchanger nil
        responder CountersResponder
      end
    end
  end
end

App.boot!

WaterDrop.setup do |water_config|
  # overwrite here any defaults or things from karafka settings
end

```

mensfeld:
Overwritable as mentioned above with better integration from 1.1

PR #226, opened 2017-10-14
Make waterdrop send_messages and raise_on_failure configurable

ankitagupta12:
### Issue
- https://github.com/karafka/karafka/issues/225

### Fix
- Define new settings inside `Karafka::Setup::Config`
- Use the settings from `config` object inside `Karafka::Setup::Configurators`

### DIFF
--- a/lib/karafka/setup/config.rb
+++ b/lib/karafka/setup/config.rb
@@ -47,6 +47,10 @@ module Karafka
       # incoming batch. It's disabled by default, not to create more objects that needed on
       # each batch
       setting :persistent, true
+      # Boolean value to define whether messages should be sent
+      setting :send_messages, true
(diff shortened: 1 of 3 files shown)


mensfeld:
@ankitagupta12 could you please namespace those settings into the ```producer``` namespace, similar to how kafka is made? Having a ```raise_on_failure``` message in a general scope is not enough descriptive as there are many failures in kafka world :)

ankitagupta12:
Hi @mensfeld. Thanks for the suggestion. I've added the settings to a namespace

mensfeld:
@ankitagupta12 will review that during the weekend. Thank you for the updates

mensfeld:
Won't merge as 1.1 copies all the common settings and the rest can be postreconfigured with waterdrop 1.0.

Issue #235, opened 2017-10-23
Rename perform to consume


mensfeld:
consume resembles Kafka ideas better then perform.

Deprecation will be added.

Issue #214, opened 2017-09-15
Expose consumer on a controller layer


mensfeld:
Now, since we have ```params``` compatibility with ruby-kafka, we can expose consumer for some additional operations (manual offset commit, etc)

mensfeld:
1.1.0.alpha2

Issue #193, opened 2017-08-18
Process stop callbacks


mensfeld:
Would be nice to add extra code that could be evaluated on startup and teardown of the app if needed.
Hooks to the runtime flow.

mensfeld:
```ruby
before_shutdown do
  current_consumer_group.commit #etc
end
```

mensfeld:
Done in https://github.com/karafka/karafka/commit/c6bbaa0759ed6db40fa0dcfa692dd342c05019e8

PR #239, opened 2017-10-27
Karafka 1.1.0

mensfeld:
- Gem bump
- Switch from Celluloid to native Thread management
- Improved shutdown process
- Introduced optional fetch callbacks and moved current ```after_received``` there as well
- Karafka will raise Errors::InvalidPauseTimeout exception when trying to pause but timeout set to 0
- Allow float for timeouts and other time based second settings
- Renamed MessagesProcessor to Processor and MessagesConsumer to Consumer - we don't process and don't consumer anything else so it was pointless to keep this "namespace"
- #232 - Remove unused ActiveSupport require
- #214 - Expose consumer on a controller layer
- #193 - Process shutdown callbacks
- Fixed accessibility of ```#params_batch``` from the outside of the controller
- connection_pool config options are no longer required
- celluloid config options are no longer required
- ```#perform``` is now renamed to ```#consume``` with warning level on using the old one (deprecated)
- #235 - Rename perform to consume
- Upgrade to ruby-kafka 0.5
- Due to redesign of Waterdrop concurrency setting is no longer needed
- #236 - Manual offset management
- WaterDrop 1.0.0 support with async
- Renamed ```batch_consuming``` option to ```batch_fetching``` as it is not a consumption (with processing) but a process of fetching messages from Kafka. The messages is considered consumed, when it is processed.
- Renamed ```batch_processing``` to ```batch_consuming``` to resemble Kafka concept of consuming messages.

Also: Gemfile points now to the github waterdrop until merged. Will update it then

@filiptepper please review as well :-)


mensfeld:
Docs will be updated once merged. Not many changes to the current once, more new features.

lib/karafka/base_controller.rb
```diff
@@ -100,18 +36,32 @@ def params_batch=(messages)
...
+    def consumer
+      Persistence::Consumer.read
+    end
+
+    # Method that will perform business logic and on data received from Kafka (it will consume
+    #   the data)
     # @note This method needs bo be implemented in a subclass. We stub it here as a failover if
     #   someone forgets about it or makes on with typo
-    def perform
+    def consume
```
mach-kernel: Great idea. This is way clearer.

lib/karafka/connection/listener.rb
```diff
@@ -31,27 +27,24 @@ def initialize(consumer_group)
...
           block.call(consumer_group.id, raw_messages)
         end
         # This is on purpose - see the notes for this method
         # rubocop:disable RescueException
       rescue Exception => e
         # rubocop:enable RescueException
         Karafka.monitor.notice_error(self.class, e)
-        @messages_consumer&.stop
-        retry if @messages_consumer
+        @consumer&.stop
```
mach-kernel: How often will this be `nil`?

lib/karafka/base_controller.rb
mensfeld: Yeah and added a deprecation warning + sidekiq-karafka supports both

lib/karafka/connection/listener.rb
mensfeld: Only if someone sends the shutdown signal before initialized (happened to me once) and in testing. But one is more than 0, that's why it go tthere in the first place.

README.md
```diff
@@ -80,39 +78,18 @@ Unfortunately, it does not yet support independent forks, however you should be
 Please run:
 
 ```bash
-bundle exec rake
+bundle exec rspec spec
```
nijikon: `bundle exec rspec`

lib/karafka/base_controller.rb
```diff
@@ -98,19 +34,27 @@ def params_batch=(messages)
     end
 
     # Executes the default controller flow, runs callbacks and if not halted
-    # will call process method of a proper backend
+    # will call perform method of a proper backend
     def call
-      run_callbacks :after_received do
-        process
-      end
+      process
```
nijikon: If it says that it will call `perform` why is it calling `process`?

lib/karafka/server.rb
```diff
@@ -4,16 +4,15 @@ module Karafka
   # Karafka consuming server class
   class Server
     class << self
-      # We need to store reference to all the consumers in the main server thread,
-      # So we can have access to them later on and be able to stop them on exit
-      attr_reader :consumers
+      # Set of consuming threads. Each consumer thread contains a single consumer
+      attr_accessor :consumer_threads
```
nijikon: Why was this changed? Where is this used?

lib/karafka/templates/karafka.rb.example
```diff
@@ -18,7 +18,7 @@ class KarafkaApp < Karafka::App
     config.kafka.seed_brokers = %w( 127.0.0.1:9092 )
     config.client_id = 'example_app'
     config.backend = :inline
-    config.batch_consuming = true
+    config.batch_fetching = true
```
nijikon: This is by default set to `true`, maybe we can remove it?

mensfeld: No because it indicates that such an option exists, so people starting with karafka are aware of it (and it is one of more important options)

lib/karafka/server.rb
mensfeld: It's not yet used directly as it only stores customer threads for reference. This will be used to #240

lib/karafka/base_controller.rb
mensfeld: Wrong method description. ```call``` calls process. ```process``` is a method for processing data that is provided by the engine, so the process is not defined on the controller, but provided by the engine. For inline it will just call ```consume``` BUT for sidekiq it behaves differently.

Issue #251, opened 2017-11-14
Shutdown upon non responding (unreachable) cluster is not possible


mensfeld:
We need a timeout after which we just kill everything and stop.

PR #263, opened 2017-11-29
#251 - Shutdown upon non responding (unreachable) cluster is not possible

mensfeld:


lib/karafka/server.rb
```diff
@@ -3,6 +3,13 @@
 module Karafka
   # Karafka consuming server class
   class Server
+    @consumer_threads = Concurrent::Array.new
+
+    # How long should we sleep between checks on shutting down consumers
+    SUPERVISION_SLEEP = 1
```
mach-kernel: I feel that this warrants a configuration key

mensfeld: @mach-kernel it just sleeps 1 second for each ```shutdown_timeout``` ```times``` to reach the shutdown_timeout timeout upon stale consumers. I don't see any benefit, especially as ```shutdown_timeout``` is configurable.

PR #264, opened 2017-11-29
Remove ActiveSupport::Inflector

ronin:
This removes dependency on `ActiveSupport::Inflector`. One class that used it was changed to use `Inflecto` instead.


lib/karafka/routing/consumer_mapper.rb
```diff
@@ -17,7 +17,7 @@ module Routing
     #   module MyMapper
     #     def self.call(raw_consumer_group_name)
     #       [
-    #         Karafka::App.config.client_id.to_s.underscope,
```
mensfeld: I guess we can't use ```underscore``` directly anymore right?

mensfeld:
Hey @ronin, thank you for your contribution. Everything looks great but just one more idea - maybe we could (while you're still on it) switch to http://dry-rb.org/gems/dry-inflector/ as we tend to use more and more dry-stuff.

lib/karafka/routing/consumer_mapper.rb
ronin: Yeah. Unless you use Karafka together with Rails application or require ActiveSupport yourself.

mensfeld: Ok, so we need to update this example as well.

mensfeld:
Great work btw. The less activesupport the better ;)

ronin:
@mensfeld Sure, I've switched to `dry-inflector` and updated the comment.

mensfeld:
Good afternoon @ronin. Could you please resolve introduced issues: https://app.coditsu.io/karafka/builds/diffs/diffs/96a6c262-783a-4c85-9d49-2c0e3912caaf/offenses/introduced - thanks :)

ronin:
@mensfeld Ok, fixed

mensfeld:
@ronin  thank you once again!

PR #277, opened 2018-01-10
Typo in docstring for Karafka::Patches::RubyKafka

michaellennox:
Assumedly `Patches` was the intended word here


mensfeld:
Hello @michaellennox. Thank you :-) for both, correcting and reading in the first place :D I think I can safely force merge this one :)

Issue #274, opened 2018-01-03
Rename controllers to consumers


mensfeld:
Controllers are confusing for a lot of Karafka users and they do not resemble by name what they are meant to do.  By renaming them to consumers, we explicitly bind them together with the data by the naming convention.

PR #279, opened 2018-01-11
#274 - Rename controllers to consumers

mensfeld:
Renames controllers to consumers to match Kafka world. #274


lib/karafka/base_consumer.rb
```diff
@@ -17,17 +17,17 @@ class BaseController
     class << self
       attr_reader :topic
 
-      # Assigns a topic to a controller and build up proper controller functionalities, so it can
+      # Assigns a topic to a consumer and build up proper consumer functionalities, so it can
```
nijikon: Assigns a topic to a consumer and build up proper consumer functionalities so that it can cooperate with the topic settings

lib/karafka/base_consumer.rb
```diff
@@ -41,14 +41,14 @@ def params_batch=(messages)
...
-    # Executes the default controller flow.
+    # Executes the default consumer flow.
     def call
       process
     end
 
     private
 
-    # We make it private as it should be accesible only from the inside of a controller
+    # We make it private as it should be accesible only from the inside of a consumer
```
nijikon: We make it private as it should be accessible only from the inside of a consumer

lib/karafka/connection/delegator.rb
```diff
@@ -20,39 +20,39 @@ def call(group_id, kafka_messages)
...
           mapped_topic_name = Karafka::App.config.topic_mapper.incoming(kafka_messages[0].topic)
           topic = Routing::Router.find("#{group_id}_#{mapped_topic_name}")
-          controller = Persistence::Controller.fetch(topic, kafka_messages[0].partition) do
-            topic.controller.new
+          consumer = Persistence::Consumer.fetch(topic, kafka_messages[0].partition) do
+            topic.consumer.new
           end
 
-          # Depending on a case (persisted or not) we might use new controller instance per each
+          # Depending on a case (persisted or not) we might use new consumer instance per each
```
nijikon: Depending on a case (persisted or not) we might use new consumer instance per each batch, or use the same one for all of them (for implementing buffering, etc.)

lib/karafka/consumers/callbacks.rb
```diff
@@ -41,7 +41,7 @@ def self.included(controller_class)
         end
       end
 
-      # Executes the default controller flow, runs callbacks and if not halted will call process
+      # Executes the default consumer flow, runs callbacks and if not halted will call process
```
nijikon: Executes the default consumer flow, runs callbacks and if not halted will call process method of a proper backend. It is here because it interacts with the default Karafka call flow and needs to be overwritten to support callbacks

lib/karafka/consumers/includer.rb
```diff
@@ -0,0 +1,51 @@
+# frozen_string_literal: true
+
+module Karafka
+  # Additional functionalities for consumers
+  module Consumers
+    # Module used to inject functionalities into a given consumer class, based on the consumer
+    # topic and its settings
+    # We don't need all the behaviors in all the cases, so it is totally not worth having
```
nijikon: We don't need all the behaviors in all the cases, so it is not worth having everything in all the cases all the time

lib/karafka/consumers/responders.rb
```diff
@@ -11,7 +11,7 @@ module Responders
       def respond_with(*data)
         Karafka.monitor.notice(self.class, data: data)
         # @note we build a new instance of responder each time, as a long running (persisted)
```
nijikon: we build a new instance of responder each time, as a long-running (persisted) consumers can respond multiple times during the lifecycle

lib/karafka/responders/builder.rb
```diff
@@ -3,30 +3,30 @@
 module Karafka
   # Responders namespace encapsulates all the internal responder implementation parts
   module Responders
-    # Responders builder is used to find (based on the controller class name) a responder that
-    # match the controller. This is used when user does not provide a responder inside routing
-    # but he still names responder with the same convention (and namespaces) as controller
+    # Responders builder is used to find (based on the consumer class name) a responder that
```
nijikon: Responders builder is used to finding (based on the consumer class name) a responder that match the consumer. We use it when user does not provide a responder inside routing, but he still names responder with the same convention (and namespaces) as consumer

lib/karafka/setup/config.rb
```diff
@@ -43,9 +43,9 @@ class Config
       # #params_batch will contain params received from Kafka (may be more than 1) so we can
       # process them in batches
       setting :batch_consuming, false
-      # Should we operate in a single controller instance across multiple batches of messages,
```
nijikon: Should we operate in a single consumer instance across multiple batches of messages, from the same partition or should we build a new one for each incoming batch. Disabling that can be useful when you want to create a new consumer instance for each incoming batch. It's disabled by default, not to create more objects that needed on each batch

PR #282, opened 2018-01-26
Monitor + performance optimizations

mensfeld:
This PR contains a new monitoring engine that tackles a lot of problems related to the previous dynamic one including:

- NewRelic support can be again enabled (with a rewrite but it will be done in a different docs related story)
- Monitor got standarized with the same format as Dry-Monitor and other (including AS Notifications) have
- Hash with indifferent is no longer supported - we will have to provide some upgrade notes and info on how you can have it, but it gives us a huge performance boost
- This PR adds support for multiple ```after_init``` hooks
- JSON parsing defaults now to string keys
- Lower memory usage due to caching optimizations
- Routing cache for faster topic detection
- Other small optimizations
- Changed ```create_time``` to ```created_at``` to follow the convention of ```received_at```


lib/karafka/base_responder.rb
```diff
@@ -75,7 +75,7 @@ class << self
       # @param options [Hash] hash with optional configuration details
       def topic(topic_name, options = {})
         self.topics ||= {}
-        topic_obj = Responders::Topic.new(topic_name, options.merge(registered: true))
+        topic_obj = Responders::Topic.new(topic_name.to_s, options.merge(registered: true))
```
nijikon: `to_s` is not needed here, we are doing that already in `Responders::Topic.new`

lib/karafka/fetcher.rb
```diff
@@ -5,47 +5,39 @@ module Karafka
...
-
-    # @return [Array<Karafka::Connection::Listener>] listeners that will consume messages
-    def listeners
-      @listeners ||= App.consumer_groups.active.map do |consumer_group|
-        Karafka::Connection::Listener.new(consumer_group)
+        # We aggregate threads here for a supervised shutdown process
+        threads.each { |thread| Karafka::Server.consumer_threads << thread }
+        threads.each(&:join)
+      # If anything crashes here, we need to raise the error and crush the runner because it means
+      # that something really bad happened
```
nijikon: that something terrible happened

lib/karafka/instrumentation/listener.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+module Karafka
+  module Instrumentation
+    # Default listener that hooks up to our instrumentation and uses its events for logging
+    # It can be removed/replaced or anyhting without any harm to the Karafka app flow
```
nijikon: It can be removed/replaced or anything without any harm to the Karafka app flow

lib/karafka/instrumentation/listener.rb
```diff
@@ -0,0 +1,85 @@
...
+        # @param event [Dry::Events::Event] event details including payload
+        def on_fetcher_call_error(event)
+          fatal "Fetcher crash due to an error: #{event[:error]}"
+        end
+
+        # Logs info about processing of a certain dataset with an inline backend
+        # @param event [Dry::Events::Event] event details including payload
+        def on_backends_inline_process(event)
+          count = event[:caller].send(:params_batch).to_a.size
+          topic = event[:caller].topic.name
```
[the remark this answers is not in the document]
mensfeld: Because what it receives as a caller is not the params object but a route object that has a different api

lib/karafka/params/params.rb
```diff
@@ -44,26 +52,27 @@ class << self
...
-            # This happens inside Kafka::FetchedProcessor
-            new(
-              parser: parser,
-              parsed: false,
-              received_at: Time.now
-            ).tap do |instance|
-              KAFKA_MESSAGE_ATTRIBUTES.each do |attribute|
-                instance[attribute] = message.send(attribute)
-              end
+          instance = new
```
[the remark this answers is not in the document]
mensfeld: nope. Blocks are expensive and we run that a loooooooooooot

lib/karafka/instrumentation/listener.rb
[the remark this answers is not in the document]
nijikon: OK, maybe a note then?

lib/karafka/params/params.rb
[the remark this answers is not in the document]
nijikon: Oh, interesting. Will measure that.

lib/karafka.rb
```diff
@@ -13,6 +13,7 @@
   dry-validation
   dry/monitor/notifications
   dry/inflector
+  dry/monitor/notifications
```
nijikon: Double require

lib/karafka/callbacks.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Karafka
+  # Additional callbacks that are used to trigger some things in given places during the
+  # system lifecycle
+  # @note Those callbacks aren't the same as consumer calbacks as they are not related to the
+  #   lifecycle of particular messages fetches but rather to the internal flow process.
+  #   They cannot be defined on a consumer callback level, because for some of those,
+  #   there aren't consumers in the memory yet and/or they aren't per consumer thread
```
nijikon: Those callbacks aren't the same as consumer callbacks as they are not related to the lifecycle of particular messages fetches but rather to the internal flow process. They cannot be defined on a consumer callback level because for some of those, there aren't consumers in the memory yet and/or they aren't per consumer thread

lib/karafka/callbacks.rb
```diff
@@ -0,0 +1,30 @@
...
+  module Callbacks
+    # Types of system callbacks that we have that are not related to consumers
+    TYPES = %i[
+      after_init
+      before_fetching
+    ].freeze
+
+    class << self
+      TYPES.each do |callback_type|
+        # Executes given callbacks set on a given moment with provided arguments
```
nijikon: Executes given callbacks set at a given moment with provided arguments

lib/karafka/callbacks/config.rb
```diff
@@ -0,0 +1,22 @@
...
+    # Additional configuration required to store procs that we will execute upon callback trigger
+    module Config
+      # Builds up internal callback accumulators
+      # @param klass [Class] Class that we extend with callback config
+      def self.extended(klass)
+        # option internal [Hash] - optional - internal karafka configuration settings that should
+        #   never be changed by users directly
+        klass.setting :internal do
+          Callbacks::TYPES.each do |callback_type|
+            # option [Array<Proc>] array of blocks that will be executed in a given moment
```
nijikon: an array of blocks that will be executed at a given moment depending on the callback type

lib/karafka/callbacks/dsl.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+module Karafka
+  module Callbacks
+    # App level dsl to define callbacks
+    module Dsl
+      Callbacks::TYPES.each do |callback_type|
+        # Allows us to define a block, that will be executed for a given momen
```
nijikon: Allows us to define a block, that will be executed for a given moment

lib/karafka/connection/listener.rb
```diff
@@ -7,15 +7,27 @@ module Connection
...
-      attr_reader :consumer_group
-
       # @param consumer_group [Karafka::Routing::ConsumerGroup] consumer group that holds details
       #   on what topics and with what settings should we listen
       # @return [Karafka::Connection::Listener] listener instance
       def initialize(consumer_group)
         @consumer_group = consumer_group
+        # This goes into the initialization, because we want to run it once even when client
+        # crashes, we don't to rerun callbacks. Also, callbacks should not interrupt with the
+        # main fetch loop
```
nijikon: This goes into the initialization because we want to run it once even when the client crashes, we don't to rerun callbacks. Also, callbacks should not interrupt with the main fetch loop

lib/karafka/instrumentation/listener.rb
```diff
@@ -9,16 +9,15 @@ class << self
         # Logs details about each received message value parsing
         # @param event [Dry::Events::Event] event details including payload
         def on_params_params_parse(event)
-          topic = event[:caller].topic
-          time = event[:time]
-          debug "Params parsing for #{topic} topic successful in #{time} ms"
+          # Keep in mind, that a caller here is a params object not a controller, so it returns
+          # a topic as a string, not a routing topic
```
nijikon: Keep in mind, that a caller here is a param object not a controller, so it returns a topic as a string, not a routing topic

spec/lib/karafka/callbacks_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+  context 'config extensions' do
+    subject(:internal) { Karafka::App.config.internal }
+
+    it { expect(internal.after_init).to be_a(Array) }
+  end
+
+  describe '#after_init' do
+    subject(:callbacks) { described_class }
+
+    before {  Karafka::App.config.internal.after_init << ->(_config) {} }
```
nijikon: Double space

Issue #284, opened 2018-01-30
Dynamic Params parent class


mensfeld:
I guess this could be pluggable so poor rails people (joke) could plug and play with indifferent access and others could appreciate the performance.

mensfeld:
Closing - done and waiting for review

PR #285, opened 2018-01-30
dynamic parent class

mensfeld:
#284


Gemfile
```diff
@@ -5,6 +5,7 @@ source 'https://rubygems.org'
 gemspec
 
 group :development, :test do
+  gem 'activesupport'
```
nijikon: `gem 'activesupport', require: false`

Issue #296, opened 2018-03-08
Instrument forcefull exit with an error


mensfeld:
When Kafka cluster is unresponsive and we decide to terminate Karafka process and it does not want to stop, upon forceful termination, we need to instrument it as an error.

PR #297, opened 2018-03-09
#296

mensfeld:


lib/karafka/instrumentation/listener.rb
```diff
@@ -86,6 +86,21 @@ def on_consumers_responders_respond_with(event)
...
 
+        # Logs info that we're going to stop the Karafka server
+        # @param _event [Dry::Events::Event] event details including payload
+        def on_server_stop(_event)
+          # We use a separate thread as logging can't be called from trap context
+          Thread.new { info "Stopping Karafka server #{::Process.pid}" }
+        end
+
+        # Logs an error that Karafka was unable to gracefully stop the server and it had to do a
+        #   force exit
```
nijikon: Logs an error that Karafka was unable to stop the server gracefully and it had to do a forced exit

PR #305, opened 2018-03-26
Rewrite callbacks to dry events

madejejej:
The end goal of this PR is to get rid of callbacks and introduce events with dry-events.

There are two types of callbacks:
1.) Home-baked for app-lifecycle events. The POC of rewriting them to dry-events is in this commit: https://github.com/karafka/karafka/commit/75737ccd6ba2d9119009df100120d42ac45baa33

2.) Activesupport based for consumer-lifecycle events `Kafka::Consumers::Callbacks`. 

We should keep backwards-compatibility with the previous behaviour. The main concern is dry-events uses payloads with `Hashes` whereas previously we relied on method signatures. 

This PR definitely needs more testing, but I prefer to get in sync early 😄

@mensfeld feedback plz 🙏 . Am I going into the right direction with this?


lib/karafka.rb
```diff
@@ -32,6 +33,11 @@ def monitor
       @monitor ||= App.config.monitor
     end
 
+    # @return [::Karafka::EventPublisher] event publisher we want to use.
+    def event_publisher
```
madejejej: should we delegate `publish` to `event_publisher`?

lib/karafka/patches/ruby_kafka.rb
```diff
@@ -17,15 +17,14 @@ def consumer_loop
                       .all
                       .values
                       .flat_map(&:values)
-                      .select { |ctrl| ctrl.respond_to?(:run_callbacks) }
```
madejejej: I have no idea why this was here. All tests are passing though 👐

lib/karafka/consumers/callbacks.rb
```diff
@@ -22,8 +22,17 @@ module ClassMethods
           # A Creates a callback wrapper
           # @param method_name [Symbol, String] method name or nil if we plan to provide a block
           # @yield A block with a code that should be executed before scheduling
-          define_method type do |method_name = nil, &block|
-            set_callback type, :before, method_name ? method_name : block
+          define_method(type) do |method_name = nil, &block|
+            register_event(type.to_s)
```
madejejej: move this to the `included` block?

lib/karafka/consumers/callbacks.rb
```diff
@@ -32,22 +41,17 @@ module ClassMethods
       def self.included(consumer_class)
         consumer_class.class_eval do
           extend ClassMethods
-          include ActiveSupport::Callbacks
-
-          # The call method is wrapped with a set of callbacks
-          # We won't run process if any of the callbacks throw abort
-          # @see http://api.rubyonrails.org/classes/ActiveSupport/Callbacks/ClassMethods.html#method-i-get_callbacks
-          TYPES.each { |type| define_callbacks type }
+          include Dry::Events::Publisher[:"karafka_#{self}_event_publisher"]
```
madejejej: should we somehow prettify the `self` here or change this symbol entirely?

It would be useful when we want to subscribe to events from `FooConsumer`, but... without knowing its class name. Doesn't sound like a valid use case.

karafka.gemspec
```diff
@@ -5,6 +5,7 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 
 require 'karafka/version'
 
+# rubocop:disable Metrics/BlockLength
```
mensfeld: sidenote: really don't like the inline disabling - will disable this for rubocop settings in coditsu (but for not it can stay)

lib/karafka/callbacks/dsl.rb
```diff
@@ -1,14 +1,29 @@
...
-        # @param [Block] block that should be executed after the initialization process
-        define_method callback_type do |&block|
-          config.callbacks.send(callback_type).push block
+      # Allows us to define a block that will be executed after Karafka has been initialized
+      # @yield [config] block that should be executed after the initialization process
+      # @yieldparam [Hash] config Karafka config
+      def after_init
+        Karafka.event_publisher.subscribe('app.after_init') do |payload|
+          # We have to unpack the payload to keep backward-compatibility
+          yield payload[:config]
```
mensfeld: we don;t need backwards compatibility outside of the  consumer callbacks. The "non" consumer callbacks backwards compatibility can be dropped :dancer:

madejejej: Are you sure? This can break someone's app on boot time. 

If we want to drop backwards-compatibility here, we could get rid of `Karafka::Callbacks` and let users subscribe to `app.after_init` and `connection.listener.before_fetch_loop` by themselves:

```ruby
# config/initializers/foo.rb
Karafka.event_publisher.subscribe('app.after_init') do |payload|
  Foo.initialize
end

# karafka.rb
require_relative 'config/initializers/foo'

class App < Karafka::App
 # ...
end

App.boot!
```

mensfeld: I am. This is targeted for the next bigger release (1.3), so there will be release notes and migration details :)

madejejej:
Benchmark code:

```ruby
class CallbackedConsumer < ApplicationConsumer
  include Karafka::Consumers::Callbacks

  after_fetch { @x = 0 }

  before_stop { @x = 1 }

  before_poll { @x = 2 }

  after_poll { @x = 3 }

  def consume
  end
end

class NonCallbackedConsumerIncludeMixin < ApplicationConsumer
  include Karafka::Consumers::Callbacks

  def consume
  end
end

class NonCallbackedConsumer < ApplicationConsumer
  def consume
  end
end

# lib/tasks/bench.rake
# frozen_string_literal: true

task :bench do
  require 'benchmark/ips'
  consumer = CallbackedConsumer.new
  non_callbacked_consumer = NonCallbackedConsumer.new
  non_callbacked_consumer_include_mixin = NonCallbackedConsumerIncludeMixin.new

  App.consumer_groups.draw do
    consumer_group :batched_group do
      batch_fetching true

      topic :non_callbacked_consumer do
        consumer NonCallbackedConsumer
        batch_consuming true
      end

      topic :non_callbacked_consumer_include_mixin do
        consumer NonCallbackedConsumerIncludeMixin
        batch_consuming true
      end

      topic :callbacked_data do
        consumer CallbackedConsumer
        batch_consuming true
      end
    end
  end

  consumer.params_batch = []
  non_callbacked_consumer.params_batch = []
  non_callbacked_consumer_include_mixin.params_batch = []

  Benchmark.ips do |x|
    x.report("With Callbacks") { consumer.call }
    x.report("With Mixin Included") { non_callbacked_consumer_include_mixin.call }
    x.report("Without Mixin Included") { non_callbacked_consumer.call }
  end
end
```

```
> ruby -v
ruby 2.5.0p0 (2017-12-25 revision 61468) [x86_64-darwin17]
```

Karafka `rewrite-callbacks-to-dry-events` branch:
```
rake bench
Warming up --------------------------------------
      With Callbacks    14.145k i/100ms
 With Mixin Included    21.116k i/100ms
Without Mixin Included
                        25.578k i/100ms
Calculating -------------------------------------
      With Callbacks    157.752k (± 5.6%) i/s -    792.120k in   5.037354s
 With Mixin Included    220.684k (± 7.0%) i/s -      1.098M in   5.002665s
Without Mixin Included
                        294.512k (± 6.6%) i/s -      1.484M in   5.059950s
```

Karafka `master` branch:

```
Warming up --------------------------------------
      With Callbacks    10.265k i/100ms
 With Mixin Included	    20.725k i/100ms
Without Mixin Included
                        25.426k i/100ms
Calculating -------------------------------------
      With Callbacks    113.819k (± 7.6%) i/s -    574.840k in   5.081517s
 With Mixin Included    228.872k (± 5.1%) i/s -      1.161M in   5.084529s
Without Mixin Included
                        283.438k (± 7.5%) i/s -      1.424M in   5.053421s
```

MacBook Pro (15-inch, Late 2016)

lib/karafka.rb
```diff
@@ -32,6 +35,11 @@ def monitor
...
```
mensfeld: can we call it just ```events```?  It's recommended by @solnic as well - then we get a nice ```events.publish``` not events_published.publish - sounds a bit railsy :D

lib/karafka/consumers/callbacks.rb
mensfeld: No - we should not subscribe to anonymous events. The only thing that worries me is the fact that it will upcase the self, right?

lib/karafka/persistence/consumer.rb
```diff
@@ -12,11 +12,16 @@ class Consumer
...
+        # @return [Hash] current thread's persistence scope hash with all the consumers
         def all
           # @note This does not need to be threadsafe (Hash) as it is always executed in a
           # current thread context
-          Thread.current[PERSISTENCE_SCOPE] ||= Hash.new { |hash, key| hash[key] = {} }
+          Thread.current[PERSISTENCE_SCOPE] ||= initial_state
+        end
+
+        # Clears current thread's persistence scope
+        def clear
```
mensfeld: why would we claer that?

mensfeld:
Massive work btw :)

lib/karafka.rb
madejejej: Good idea!

I'm also wondering whether migrating to dry-events isn't an overkill in case of these callbacks. These are only 2 events in the whole lifecycle of the application.

Are there any other use-cases beyond setting-up 3rd party libraries here?

lib/karafka/consumers/callbacks.rb
madejejej: It will, and if the `self` is namespaced the namespace will be here too. If we don't want to subscribe to publishers by magic strings this shouldn't be an issue though.

mensfeld:
@madejejej once you're back from your trips, could you: 

1. Retarget to 1.3-wip branch
2. Fix https://app.coditsu.io/karafka/builds/commit_builds/a757367b-41f6-4b6c-a3ac-3ae1c3c9c79a/validations

I guess I can apply some more remarks but those two things need to be done either way.

madejejej:
@mensfeld Retargeted the branch.

The YARD violation is unremovable:

`Raised when an object is recognized but cannot be documented. This generally occurs when the Ruby syntax used to declare an object is too dynamic in nature.`

and it's because dry-events uses the module builder pattern.

I removed the other ones.

lib/karafka.rb
```diff
@@ -11,9 +11,12 @@
   require_all
   dry-configurable
   dry-validation
+  dry/events/publisher
   dry/inflector
   dry/monitor/notifications
-  active_support/callbacks
+  active_support/core_ext/class/attribute
+  active_support/core_ext/hash/except
```
mensfeld: do we need this? I believe I removed it recently (guess in 1.2.2). Could you please verify? We aim to remoe AS at all.

lib/karafka.rb
```diff
@@ -32,6 +35,11 @@ def monitor
       @monitor ||= App.config.monitor
     end
 
+    # @return [::Karafka::EventPublisher] event publisher we want to use.
+    def events
+      @events ||= EventPublisher.new
```
mensfeld: Could you add following note to the yard doc:

We don't need to ensure that it is not executed in a thread for the first time (assingment is not thread-safe) as it gets used during the non-threaded boot process.

lib/karafka/app.rb
```diff
@@ -15,7 +14,7 @@ class << self
       def boot!
         Setup::Config.validate!
         Setup::Config.setup_components
-        Callbacks.after_init(Karafka::App.config)
+        events.publish('app.after_init', config: Karafka::App.config)
```
mensfeld: we should have a single place where all the system events are defined (apart from the dynamic per controller once). This makes code easier to read for users.

lib/karafka.rb
mensfeld: I guess we could also use their aproach from the docs: http://dry-rb.org/gems/dry-events/ - in terms of the API usage and event registration. Especially as I requested to raise an error on non registered events.

lib/karafka.rb
madejejej: Still needed:

    karafka/lib/karafka/base_responder.rb:63:in `<class:BaseResponder>': undefined method `class_attribute' for Karafka::BaseResponder:Class (NoMethodError)

    karafka/lib/karafka/setup/configurators/water_drop.rb:16:in `block in setup': undefined method `except' for #<Hash:0x00007fce633972c0> (NoMethodError)

We could remove/provide implementations for those but probably this PR isn't the best place to do it.

lib/karafka.rb
madejejej: Could you be more specific? Are you suggesting removing the `EventPublisher` class and including the Publisher mixin in `Karafka::App` instead?

lib/karafka/app.rb
madejejej: Again not sure what you mean here. They are defined in the `EventPublisher`. We could move the event names to constants defined in a separate module, but since you requested that dry-events should raise an error when publishing an unregistered event, I'm not sure we need to.

lib/karafka.rb
mensfeld: You're absolutely right :) Thank you for the clarification.

lib/karafka.rb
mensfeld: Giving it a second though. Please ignore.

lib/karafka/app.rb
mensfeld: Dunno why but during the review I completely missed this: https://github.com/karafka/karafka/pull/305/files#diff-9956e073809cb05c57c6dd8df8777aeeR8 - everything is fine here apart from moving it to a constant just like events for the instrumentation: https://github.com/karafka/karafka/blob/master/lib/karafka/instrumentation/monitor.rb#L23

mensfeld:
@madejejej remved the WIP for now. Also would be good to update with 1.3 when you will be working with it :)

lib/karafka/consumers/callbacks.rb
```diff
@@ -19,11 +19,23 @@ module Callbacks
...
         TYPES.each do |type|
-          # A Creates a callback wrapper
+          # Creates a callback wrapper
+          #
           # @param method_name [Symbol, String] method name or nil if we plan to provide a block
           # @yield A block with a code that should be executed before scheduling
-          define_method type do |method_name = nil, &block|
-            set_callback type, :before, method_name ? method_name : block
+          define_method(type) do |method_name = nil, &block|
+            key = "consumers.#{Helpers::Inflector.underscore(self)}.#{type}"
```
madejejej: this works weird:

```
2.5.1 :003 > Karafka::Helpers::Inflector.underscore('ModuleA::ModuleB::ModuleC::MyConsumer')
 => "module_a_module_b/module_c/my_consumer"
```

madejejej: put the key to a method

lib/karafka/helpers/inflector.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module Karafka
+  module Helpers
+    module Inflector
+      ENGINE = Dry::Inflector.new
+
+      def self.underscore(element)
+        ENGINE.underscore(element.to_s).sub('/', '_')
```
madejejej: I think this should be `gsub` instead of `sub`. Currently:

```
2.5.1 :003 > Karafka::Helpers::Inflector.underscore('ModuleA::ModuleB::ModuleC::MyConsumer')
 => "module_a_module_b/module_c/my_consumer"
```

lib/karafka.rb
```diff
@@ -39,7 +39,7 @@ def monitor
     # @note We don't need to ensure that it is not executed in a thread for the first time
     # (assingment is not thread-safe) as it gets used during the non-threaded boot process.
```
mensfeld: typo

mensfeld:
@madejejej I've fixed the visibility problems. Feel free to benchmark.

madejejej:
@mensfeld benchmarks

For the current version:

```
Warming up --------------------------------------
      With Callbacks     6.483k i/100ms
 With Mixin Included     6.220k i/100ms
Without Mixin Included
                        28.224k i/100ms
Calculating -------------------------------------
      With Callbacks     67.726k (±12.2%) i/s -    337.116k in   5.053245s
 With Mixin Included     62.532k (± 8.8%) i/s -    311.000k in   5.014042s
Without Mixin Included
                        303.951k (± 9.5%) i/s -      1.524M in   5.059467s
```

My last commit (10a11af)
 
```
Warming up --------------------------------------
      With Callbacks    16.476k i/100ms
 With Mixin Included    24.041k i/100ms
Without Mixin Included
                        30.501k i/100ms
Calculating -------------------------------------
      With Callbacks    178.292k (± 3.7%) i/s -    906.180k in   5.089617s
 With Mixin Included    260.591k (± 3.9%) i/s -      1.322M in   5.082329s
Without Mixin Included
                        345.659k (± 2.9%) i/s -      1.739M in   5.034322s
```

Looks like commit bdf7ff8 introduced slowed everything down:

```
Warming up --------------------------------------
      With Callbacks     7.318k i/100ms
 With Mixin Included     6.524k i/100ms
Without Mixin Included
                        31.409k i/100ms
Calculating -------------------------------------
      With Callbacks     78.455k (± 3.7%) i/s -    395.172k in   5.044549s
 With Mixin Included     66.720k (±10.5%) i/s -    332.724k in   5.049827s
Without Mixin Included
                        332.063k (± 6.9%) i/s -      1.665M in   5.039372s
```


Here's the benchmark source: https://gist.github.com/madejejej/a9956ce3955cf922ccff83e33edc9483

mensfeld:
That is interesting. I will look into this shortly after I'm back from Japan.

madejejej:
https://www.youtube.com/watch?v=DdWZKb659K0

mensfeld:
@madejejej looked into this now. The results aren't that far now:

```
# Your last comment
With Callbacks    207.215k (± 6.5%) i/s -      1.033M in   5.008800s
With Mixin Included    314.171k (± 6.0%) i/s -      1.592M in   5.085102s
Without Mixin Included
                        419.958k (± 5.3%) i/s -      2.134M in   5.095775s

# My new local changes
With Callbacks    180.682k (± 2.5%) i/s -    904.233k in   5.007624s
With Mixin Included    234.302k (±12.4%) i/s -      1.148M in   5.006763s
Without Mixin Included
                        443.658k (± 7.1%) i/s -      2.231M in   5.060608s
```

mensfeld:
@madejejej could you also rerun it on your hardware? I did some small optimizations and the results are much better.

madejejej:
@mensfeld it's still around 30% slower than the previous version. What's funny is that it's a bit faster than the ActiveSupport version when the callbacks are running, but slower if we only include the mixin 😆 

```
Warming up --------------------------------------
      With Callbacks    13.752k i/100ms
 With Mixin Included    19.133k i/100ms
Without Mixin Included
                        33.506k i/100ms
Calculating -------------------------------------
      With Callbacks    148.305k (± 4.1%) i/s -    742.608k in   5.015653s
 With Mixin Included    200.209k (± 5.9%) i/s -      1.014M in   5.083639s
Without Mixin Included
                        353.321k (± 9.2%) i/s -      1.776M in   5.077140s
```

lib/karafka/consumers/callbacks.rb
```diff
@@ -24,7 +24,7 @@ module ClassMethods
           # @param method_name [Symbol, String] method name or nil if we plan to provide a block
           # @yield A block with a code that should be executed before scheduling
           define_method(type) do |method_name = nil, &block|
-            key = "consumers.#{Helpers::Inflector.underscore(self)}.#{type}"
```
madejejej: I'd even try memoizing whole strings here and in the `call` method.

mensfeld:
Hmmm,

Current head on this work:

```
Calculating -------------------------------------
      With Callbacks    148.305k (± 4.1%) i/s -    742.608k in   5.015653s
 With Mixin Included    200.209k (± 5.9%) i/s -      1.014M in   5.083639s
Without Mixin Included
                        353.321k (± 9.2%) i/s -      1.776M in   5.077140s
```

AS base (current master):

```
Warming up --------------------------------------
      With Callbacks     6.483k i/100ms
 With Mixin Included     6.220k i/100ms
Without Mixin Included
                        28.224k i/100ms
Calculating -------------------------------------
      With Callbacks     67.726k (±12.2%) i/s -    337.116k in   5.053245s
 With Mixin Included     62.532k (± 8.8%) i/s -    311.000k in   5.014042s
Without Mixin Included
                        303.951k (± 9.5%) i/s -      1.524M in   5.059467s
```

To me it seems, that it's 2-3 times faster than the current master for each case.

mensfeld:
@madejejej you think you could resolve the conflicts?

madejejej:
@mensfeld done!

madejejej:
@mensfeld I think we still can cache more strings:

```
KARAFKA_ROOT_DIR=. KARAFKA_BOOT_FILE=bench.rb ruby bench.rb
Warming up --------------------------------------
      With Callbacks    13.506k i/100ms
 With Mixin Included    19.909k i/100ms
Without Mixin Included
                        32.211k i/100ms
Calculating -------------------------------------
      With Callbacks    143.816k (± 2.8%) i/s -    729.324k in   5.075314s
 With Mixin Included    178.659k (±16.8%) i/s -    856.087k in   5.033599s
Without Mixin Included
                        352.450k (± 4.0%) i/s -      1.772M in   5.036113s
```

mensfeld:
@madejejej would you be willing to implement that or should I do it?

mensfeld:
Improving the cache won't be so easy...

mensfeld:
@madejejej correct me if I'm wrong, but it seems, that the only case that is slower, is the ` With Mixin Included` for which callbacks / events aren't being used. That is to say - you include a mixin that you're not going to use and you get a penalty. If it is correct, we can merge it the way it is as it is faster for any other case.

karafka.gemspec
nijikon: We should drop this disable, `Metrics/BlockLength` should not be run for `*.gemspec`

lib/karafka/helpers/inflector.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Karafka
```
nijikon: Please document this class

lib/karafka/persistence/consumer.rb
nijikon: I don't like adding methods that are only used in specs. Can there be a scenario that this method will be used in production environment? If not, can we move it to specs?

spec/lib/karafka/connection/listener_spec.rb
```diff
@@ -16,7 +16,9 @@
     let(:client) { listener.send(:client) }
 
     it 'expects to run callbacks and start the main fetch loop' do
-      expect(Karafka::Callbacks).to receive(:before_fetch_loop).with(consumer_group, client)
+      expect(Karafka.events).to receive(:publish).with('before_fetch_loop',
```
nijikon:
```ruby
expect(Karafka.events).to receive(:publish)
  .with('before_fetch_loop', consumer_group: consumer_group, client: client)
```

spec/lib/karafka/instrumentation/event_publisher_spec.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+RSpec.describe Karafka::Instrumentation::EventPublisher do
```
nijikon: Can we add specs?

madejejej:
@mensfeld it's faster than the current master, but around 15-20% slower than 10a11af7

But yeah, I think we should merge it, as it's a big improvement nevertheless.

mensfeld:
There's always a tradeoff ;) - I will apply the quality remarks and merge it during the weekend.

mensfeld:
Refactored a bit. Now looks awesome!

nijikon:
💥

mensfeld:
:fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks: :fireworks:

mensfeld:
The longest running open PR in the Karafka ecosystem ever :) Great work @madejejej!

Issue #306, opened 2018-04-04
Using file logger from within a trap context upon shutdown is impossible


mensfeld:
If I subscribe to ```server.stop``` and ```server.stop.error``` and I try to use logger from within, Karafka throws an error:

```ruby
ThreadError: can't be called from trap context
```

fix idea:

Broadcast notification from within a separate thread for supervisioned shutdown elements
```ruby
 Thread.new { Karafka.monitor.instrument('server.stop', {}) }.join
```

PR #307, opened 2018-04-04
#306

mensfeld:


lib/karafka/server.rb
```diff
@@ -67,7 +67,12 @@ def start_supervised
       # Stops Karafka with a supervision (as long as there is a shutdown timeout)
       # If consumers won't stop in a given timeframe, it will force them to exit
       def stop_supervised
-        Karafka.monitor.instrument('server.stop', {})
+        # Due to the fact that this is called in the trap context, there is a chance that
+        # instrumentation listeners contain things that aren't allowed from within a trap
+        # context. To bypass that (instead of telling users not to do things they need to)
+        # we just spin up a thread ti instrument server.stop and server.stop.error and wait
+        # until they're finished
```
nijikon: Because this is called in the trap context, there is a chance that instrumentation listeners contain things that aren't allowed from within a trap context. To bypass that (instead of telling users not to do things they need to) we spin up a thread to instrument server.stop and server.stop.error and wait until they're finished

PR #308, opened 2018-04-04
cleanup ruby versions

nijikon:


mensfeld:
I guess it's ok. No major changes should be introduced in the in-between versions.

PR #338, opened 2018-05-19
#331

### LINKED ISSUE #331: Disallow building groups without topics
```ruby
consumer_group :whatever do
end
```

should not be possible.

mensfeld:
closes #331


lib/karafka/routing/builder.rb
```diff
@@ -27,7 +27,7 @@ def draw(&block)
         each do |consumer_group|
           hashed_group = consumer_group.to_h
           validation_result = Karafka::Schemas::ConsumerGroup.call(hashed_group)
-          return if validation_result.success?
```
nijikon: La bug 🗡

mensfeld: Yeah, don't even tell me ;)

PR #349, opened 2018-05-25
issue-341

### LINKED ISSUE #341: Split connection delegator into batch delegator and single_delegator
This way we can much easier manage the delegation to consumers and cleaner provide support for headers.

mensfeld:
closes #341

### DIFF
--- a/lib/karafka/connection/listener.rb
+++ b/lib/karafka/connection/listener.rb
@@ -16,19 +16,13 @@ module Karafka
 
       # Runs prefetch callbacks and executes the main listener fetch loop
       def call
-        Karafka::Callbacks.before_fetch_loop(
-          @consumer_group,
-          client
-        )
+        Karafka::Callbacks.before_fetch_loop(@consumer_group, client)
         fetch_loop
       end
 
       private
 
       # Opens connection, gets messages and calls a block for each of the incoming messages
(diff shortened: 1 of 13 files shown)


lib/karafka/connection/listener.rb
```diff
@@ -16,30 +16,26 @@ def initialize(consumer_group)
...
       #   Since it is run inside Karafka::Connection::ActorCluster - catching all the exceptions
       #   won't crash the whole cluster. Here we mostly focus on catchin the exceptions related to
       #   Kafka connections / Internet connection issues / Etc. Business logic problems should not
       #   propagate this far
       def fetch_loop
-        client.fetch_loop do |raw_messages|
-          # @note What happens here is a delegation of processing to a proper processor based
-          #   on the incoming messages characteristics
-          Karafka::Connection::Delegator.call(@consumer_group.id, raw_messages)
+        delegators = { message: MessageDelegator, batch: BatchDelegator }
```
nijikon: Move this to constant maybe?

mensfeld: Yes and no. If I move it to constants, I will have to explicitly manage the internal load order to load delegators before the listener. Didn't decide to do that, since listener is invoked once per process lifetime.

PR #353, opened 2018-05-28
reorder config comments to match options with docs

filiptepper:


nijikon:
@mensfeld I think that we should have it ordered alphabetically in both here and the docs. What do you think?

mensfeld:
@nijikon I don't mind renaming the karafka section for configs to be alphabetical but this "random" order within Kafka section is not random at all. It matches Ruby-kafka settings declarations in various places and helps me track changes and stay in sync with proper argument and options injection through ruby-kafka client lifetime.

See for example:

https://github.com/karafka/karafka/blob/master/lib/karafka/setup/config.rb#L80
and https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/client.rb#L284

The order is the same.

nijikon:
I understand your argument here, but now we have `pause_timeout` which is not in that file. So how is that to the whole concept?

mensfeld:
@nijikon yes. The `pause_timeout` should be removed from this section and moved out. Feel free to check the orders of particular `kafka.` namespace options and match them with corresponding Karafka options to get them in proper (matching) order in a separate PR :). Thanks in advance!

PR #357, opened 2018-06-01
Fix a wrong file permission

chiastolite:
I found wrong permissions on some files by following command, and fix them.

`find . -perm 755 -type f | grep -v '.git/' | grep -v 'bin/'`


mensfeld:
Thank you for your contribution. We will review it shortly. @nijikon could we incorporate that types of checks within Coditsu to ensure, that this won't happen again?

nijikon:
@mensfeld Yes, we can incorporate this check to Coditsu, we can also generate a command in the validator that would allow a person to fix this issues.

mensfeld:
Ok. Will merge. Will add that to Coditsu once back from Japan as well.

PR #361, opened 2018-06-01
Single params consuming mode automatically parses data spec

### LINKED ISSUE #360: Single params consuming mode automatically parses data specs
Outdated:

When running in a single mode, the consumer automatically unparses the data for params, which should not happen as the rejection based on metadata should still work for single params handling.

Justification:

Actually this behavior is really consistent for both `#first` and `#last` and for single consumption we should use exactly the same API as for batched for unparsed data: `#params_batch.to_a`.

I will update spec accordingly.

mensfeld:
Closes #360


mensfeld:
Reason for this PR has changed. Please see corresponding issue.

Issue #375, opened 2018-07-18
1.3-wip Sidekiq Backend parser mismatch


mensfeld:
Due to the fact, that parser is assigned before merging the serialized message, there can be a case where parser is overwritten due to a payload.

PR #376, opened 2018-07-18
Fix for parser mismatch

mensfeld:
close: https://github.com/karafka/karafka/issues/375


nijikon:
@mensfeld Nice! Another `active_support` dependency of the list 🎉

PR #378, opened 2018-07-18
Remove persistence in favor of persisting

mensfeld:
This will allow us to optimize how consumers are being built and pre heat everything.

It will also simplify a bit the codebase and allow further steps towards Ruby 3.0 support.


nijikon:
@mensfeld I think that we could introduce a warning if anybody is still setting it.

mensfeld:
@nijikon As always I will provide all the migration guides.

Issue #408, opened 2018-09-25
Responder Topic Lookup Bug on Heroku


dafeiner:
Original error:
```ruby
OrderResponder.call Order.last
```
```
Traceback (most recent call last):
        1: from (irb):25
NoMethodError (undefined method `async?' for nil:NilClass)
```

```ruby
# order_responder.rb
class OrderResponder < ApplicationResponder
  topic :test

  def respond(order)
    respond_to :test, order
  end
end

# karafka.rb

class PrefixTopicMapper
  def initialize(prefix)
    @prefix = prefix
  end

  def incoming(topic)
    topic.to_s.gsub(@prefix, '')
  end

  def outgoing(topic)
    "#{@prefix}#{topic}"
  end
end

class KarafkaApp < Karafka::App
  setup do |config|
    # other setup stuff
    config.topic_mapper = PrefixTopicMapper.new(ENV['KAFKA_PREFIX'])
  end
end

```

It looks like this was broken in #373 ...

When I initialize a new OrderResponder in irb

```
irb(main):010:0> r = OrderResponder.new
=> #<OrderResponder:0x000000000acb1d08 @parser_class=Karafka::Parsers::Json, @messages_buffer={}>
```

Then call the respond_to method directly and look at the messages buffer the topic is namespaced with the prefix (https://github.com/karafka/karafka/blob/b10c4acd2cde4dbf95c524de5088856b5bf1bba0/lib/karafka/base_responder.rb#L194):

```ruby
r.send :respond_to, :test, Order.last
ap r.messages_buffer
{
    "test" => [
        [0] [
            [0] # Order Data,
            [1] {
                :topic => "prefix-123.test"
            }
        ]
    ]
}
```

But the responder topics hash is keyed without the prefix (https://github.com/karafka/karafka/blob/b10c4acd2cde4dbf95c524de5088856b5bf1bba0/lib/karafka/base_responder.rb#L78):

```ruby
irb(main):045:0> ap r.topics
{
    "test" => #<Karafka::Responders::Topic:0x0000000009dce098 @name="test", @options={:registered=>true}>
}
```

So there is a disconnect there. Internally we probably don't want to use the mapped topic name until the very end when we pass it to the `WaterDrop` producer.

I may make a PR later, but it looks like we are at the minimum missing a spec here with prefixed topic names.

mensfeld:
Hello @dafeiner. Thank you for reporting this and really sorry, that you've encountered problems working with Karafka.

[...]

PR #409, opened 2018-09-25
Fix internal use of mapped topic name

dafeiner:
The @messages_buffer attribute had the external topic name stored in it but the responder `#topics` hash was using the internal (easily readable) non mapped version. Only right at the moment when we communicate with Kafka should we re-map the topic name.

See issue https://github.com/karafka/karafka/issues/408 for more details


dafeiner:
@mensfeld fyi

mensfeld:
Thank you for this PR. On a first glance, it looks good. I will the CI soon for master and once that's done, will take care of it and make a release for 1.2 branch.

Issue #423, opened 2018-10-11
Lower default kafka.fetcher_max_queue_size to 10


mensfeld:
It should be set to 10 not 100 as 100 is consuming a loooot of memory and I would rather go with a safe setting.

PR #424, opened 2018-10-15
#423

mensfeld:
close https://github.com/karafka/karafka/issues/423


mensfeld:
Rejected due to a duplication.

Issue #428, opened 2018-10-18
remove offset_retention_time


mensfeld:
This setting is for brokers and does not need to be exposed. Also it is not present for other drivers apart from ruby-kafka.

PR #430, opened 2018-10-19
drop offset_retention_time config option

nijikon:
Closes https://github.com/karafka/karafka/issues/428


mensfeld:
Closing it for now. It seems to be used by some of the people and until we switch to ruby-kafka, this will be on hold (we will get back to that one).

PR #444, opened 2019-01-03
add certificate and private key validation

### LINKED ISSUE #398: Validate certs strings format upon entry
If there is a way to detect, that the cert string is not valid (missing chars, etc) we should catch it and handle.

nijikon:
Closes #398


mensfeld:
@nijikon could you check why the CI is failing? Thanks!

mensfeld:
Great job!

Issue #460, opened 2019-01-25
Decouple responder "parser" (generator?) from topic.parser


benissimo:
Add an option to override the (built-in, currently hard-coded) logic that assumes a responder uses the same "parser" as the consumer. (That was already a questionable assumption even if you're just dealing with JSON and XML. With Avro, hard-coding a parser won't work as the responder typically writes to a different topic, and even if consumer and responder topics both use Avro, it's unlikely they use the same schema.)

Add support for the following cases:

1)
Consumer uses one parser, responder uses another (consumer parses JSON, responder generates XML, etc.). Allow a consumer to specify its own default format.

2)
Responder generates messages with different formats depending on the topic it responds to.
A single responder can respond to multiple topics. They might not all share the same format. So allow `respond_to` to pass in an optional parser.

mensfeld:
I again agree it was a bad assumption on our (primarily mine) side.

mensfeld:
@benissimo would you mind looking at this as well? https://github.com/karafka/karafka/pull/464 once this is merged, I will close this issue. Thank you for your help! This is a really good change that makes things much cleaner!

mensfeld:
Thank you for your contribution! The idea of splitting that was awesome and the more I think about it, the more the previous approach feels bad!

PR #461, opened 2019-01-25
Override generator (parser) for responder

benissimo:
This addresses the two cases listed on https://github.com/karafka/karafka/issues/460

Case 1:

```
Class MyResponder < BaseResponder
  # Takes precedence over the consumer's parser
  def default_generator_class
    MyCustomParser
  end
end
```

Case 2:

```
Class MyResponder < BaseResponder

  def respond(data)
    if some_condition
      # Override each time you call respond_to()...
      respond_to :topic1, data, generator: MyCustomParser
    else
      respond_to :topic2, data, generator: MyOtherCustomerParser
    end

    respond_to :topic3, data, generator: YetAnotherCustomerParser
  end
end

```

I struggled with naming, feedback welcome. "Parser" is Karafka's term
for what's known in Kafka as a deserializer. It seems confusing to
call the serializer a parser too. Since the method called on it is
`generate()` I chose "generator".

### DIFF
--- a/lib/karafka/base_responder.rb
+++ b/lib/karafka/base_responder.rb
@@ -96,8 +96,8 @@ module Karafka
     # @param parser [Object] parser that we can use to generate appropriate string
     #   or nothing if we want to default to Karafka::Parsers::Json
     # @return [Karafka::BaseResponder] base responder descendant responder
-    def initialize(parser = Karafka::App.config.parser)
-      @parser = parser
+    def initialize(consumer_topic_parser = Karafka::App.config.parser)
+      @consumer_topic_parser = consumer_topic_parser
       @messages_buffer = {}
     end
 
@@ -115,6 +115,10 @@ module Karafka
       validate_usage!
       validate_options!
       deliver!
+    end
+
+    def generator_class
+      default_generator_class || @consumer_topic_parser
     end
 
     private
@@ -183,18 +187,25 @@ module Karafka
     # as many times as we need. Especially when we have 1:n flow
     # @param topic [Symbol, String] topic to which we want to respond
     # @param data [String, Object] string or object that we want to send
-    # @param options [Hash] options for waterdrop (e.g. partition_key)
+    # @param options [Hash] options for waterdrop (e.g. partition_key) and a generator override.
     # @note Respond to does not accept multiple data arguments.
     def respond_to(topic, data, options = {})
       # We normalize the format to string, as WaterDrop and Ruby-Kafka support only
       # string topics
       topic = topic.to_s
 
+      generator = options.delete(:generator) { generator_class }
+
       messages_buffer[topic] ||= []
       messages_buffer[topic] << [
-        @parser.generate(data),
+        generator.generate(data),
         options.merge(topic: topic)
       ]
+    end
+
+    # Override this method in a subclass to set a default parser for your responder.
+    # Otherwise, responder falls back to using the same parser as the consumer.
+    def default_generator_class
     end
 
     # @param options [Hash] options for waterdrop
--- a/lib/karafka/base_responder.rb
+++ b/lib/karafka/base_responder.rb
@@ -48,6 +48,17 @@ module Karafka
   #     end
   #   end
   #
+  # @example Specify serializer for a topic
+  #   class Responder < BaseResponder
+  #     topic :xml_topic, serializer: MyXMLSerializer
+  #
+  #     def respond(data)
+  #       data.each do |subset|
+  #         respond_to :xml_topic, subset
+  #       end
+  #     end
+  #   end
+  #
   # @example Accept multiple arguments to a respond method
   #   class Responder < BaseResponder
   #     topic :users_actions
@@ -117,8 +128,8 @@ module Karafka
       deliver!
     end
 
-    def generator_class
-      default_generator_class || @consumer_topic_parser
+    def serializer(topic)
+      self.class.topics[topic].serializer || @consumer_topic_parser
     end
 
     private
@@ -187,25 +198,19 @@ module Karafka
     # as many times as we need. Especially when we have 1:n flow
     # @param topic [Symbol, String] topic to which we want to respond
     # @param data [String, Object] string or object that we want to send
-    # @param options [Hash] options for waterdrop (e.g. partition_key) and a generator override.
+    # @param options [Hash] options for waterdrop (e.g. partition_key).
     # @note Respond to does not accept multiple data arguments.
     def respond_to(topic, data, options = {})
+      serializer = serializer(topic)
       # We normalize the format to string, as WaterDrop and Ruby-Kafka support only
       # string topics
       topic = topic.to_s
 
-      generator = options.delete(:generator) { generator_class }
-
       messages_buffer[topic] ||= []
       messages_buffer[topic] << [
-        generator.generate(data),
+        serializer.generate(data),
         options.merge(topic: topic)
       ]
-    end
-
-    # Override this method in a subclass to set a default parser for your responder.
-    # Otherwise, responder falls back to using the same parser as the consumer.
-    def default_generator_class
     end
 
     # @param options [Hash] options for waterdrop
--- a/lib/karafka/base_responder.rb
+++ b/lib/karafka/base_responder.rb
@@ -104,7 +104,7 @@ module Karafka
     attr_reader :messages_buffer
 
     # Creates a responder object
-    # @param parser [Object] parser that we can use to generate appropriate string
+    # @param consumer_topic_parser [Object] parser that we can use to generate appropriate string
     #   or nothing if we want to default to Karafka::Parsers::Json
     # @return [Karafka::BaseResponder] base responder descendant responder
     def initialize(consumer_topic_parser = Karafka::App.config.parser)
@@ -128,6 +128,8 @@ module Karafka
       deliver!
     end
 
+    # @param topic [Symbol, String] topic to which we want to respond
+    # @return [Object, Class] serializer for serialization of the outgoing data per topic
     def serializer(topic)
       self.class.topics[topic].serializer || @consumer_topic_parser
     end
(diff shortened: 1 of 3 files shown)


mensfeld:
Wow! That was fast. I will take a look next week if you don't mind. Same @nijikon 

Coditsu won't run on the external PR (Travis limitation).

mensfeld:
Yeah: I will have to fix travis setup not to run it as a dependent pipeline (or silence the coditsu container) @nijikon : https://travis-ci.org/karafka/karafka/jobs/484444819

Anyhow @benissimo nothing you have to worry about.

benissimo:
> Wow! That was fast. I will take a look next week if you don't mind. Same @nijikon
> 
> Coditsu won't run on the external PR (Travis limitation).

Thanks, another approach could be to map topics to parser/generators. (It seems safe to assume that each topic uses one and only one format.) But that would require a larger rewrite of Karafka: we could add responder topics to the list of routes assigned in karafka.rb:

```
  consumer_groups.draw do
    batch_fetching true

    topic :topic1 do
      consumer MyConsumer1
      parser MyParser1
    end

   # Register topic 2 even if we have _no_ consumer for it, so that any responder writes to it will use MyParser
    topic :topic2 do
      parser MyParser2
    end
```
but that then complicates the structure of drawing consumer groups. topic2 doesn't necessarily belong to any consumer, it might simply be used by one or more responders. Having to define it here, nested within a consumer group, is confusing (if topic2 is written to but not read by any consumers in a given group, in what sense does topic2 "belong to" that group?). Another idea might be to move the mapping of topic -> parser elsewhere, outside of the assigning of consumer groups and consumers. I erred on the side of making the smallest change possible to satisfy the cases mentioned in https://github.com/karafka/karafka/issues/460
but a more sweeping change might be a better longterm move. Unless there's a use case for sending varying formats to the same topic (I can't think of one myself but maybe one exists?), binding parsers to topics seems a better fit.

mensfeld:
> we could add responder topics to the list of routes assigned in karafka.rb

Actually, responders will be moved out of Karafka. They will be an optional plugin via dry-core. Responders aren't relevant to the consumption process, thus would prefer not to inject those in the routing.

>  Unless there's a use case for sending varying formats to the same topic (I can't think of one myself but maybe one exists?), binding parsers to topics seems a better fit.

I agree. Even if there's such a case, the default parser won't do anything with a string (see: https://github.com/karafka/karafka/blob/master/lib/karafka/parsers/json.rb#L32), so you can just use the default one and preparse your message before it hits the framework.

mensfeld:
About the naming: we could switch the `parser` name to `deserializer` and the `generator` to `serializer` on a per responder basis.

This would not require much work: you would have to retarget this PR to `1.3-wip`, that way it could go as a breaking change with the incoming Karafka version and I could take care of the deserialization. The `parser` name is misleading and it shouldn't hold both the responsibilities.

mensfeld:
Actually when I think about it, we could do something like that for 1.3 (and extract that as a plugin later on):

```ruby
consumer_groups.draw do
  batch_fetching true

  topic :topic1 do
    consumer MyConsumer1
    # This would replace the parser we know
    deserializer Karafka::Serialization::JsonDeserializer
    # This would be used in the responder with a fallback to JsonDeserializer
    serializer Karafka::Serialization::AvroSerializer
  end
end
```

we could extract this routing option later on into the responders plugin but for 1.3 anyhow, responders will stay in the karafka-core.

The only significant change would be here: https://github.com/karafka/karafka/blob/1.3-wip/lib/karafka/consumers/responders.rb#L19

Instead of passing the parser, we would pass the serializer.

benissimo:
> Actually when I think about it, we could do something like that for 1.3 (and extract that as a plugin later on):
> 
> ```ruby
> consumer_groups.draw do
>   batch_fetching true
> 
>   topic :topic1 do
>     consumer MyConsumer1
>     # This would replace the parser we know
>     deserializer Karafka::Serialization::JsonDeserializer
>     # This would be used in the responder with a fallback to JsonDeserializer
>     serializer Karafka::Serialization::AvroSerializer
> ```

^ This should only be used when serializing messages to the "topic1" topic. But a responder may write to (many) other topics. It would be best to tie serialization to each topic. Is that what you had in mind here? If so consider revising the inline comment as something like:
```
# This would be used by any responders that write to topic1.
```

>   ```ruby
>   end
> end
> ```
> 
> we could extract this routing option later on into the responders plugin but for 1.3 anyhow, responders will stay in the karafka-core.
> 
> The only significant change would be here: https://github.com/karafka/karafka/blob/1.3-wip/lib/karafka/consumers/responders.rb#L19
> 
> Instead of passing the parser, we would pass the serializer.

I like your idea of changing the names ("serializer/deserializer" instead of "parser"). I'm still not sure about specifying topic serializers in this section where we're defining routes.  How topics are fetched/consumed seems distinct from how topics are serialized/deserialized. Fetching/consuming a topic is done by a consumer, which naturally belongs to a single consumer group. Serializing should belong to a topic and be a common behavior across all writes to a topic: nothing needs to prevent a topic from being written to by multiple consumers, belonging to different consumer groups. (For example multiple consumers might all write to an "error" or "dead-letter" topic.) But the routing code assumes (understandably) that each topic belongs to a single consumer group. That might be true in the case of consumers (though even there, not necessarily) but it seems a poor fit for responders.

In any case, however serializers/deserializers are assigned to topics, it might be useful if there's a way to customize them on a per-consumer/responder basis. For example I might have a `topic1` that has an `AvroSerDe1` (SerDe = Serializer/Deserializer) class but I might occasionally write to it using an `AvroSerDe2` class (supposing that a field which is optional in `AvroSerDe1` is required in `AvroSerDe2`, for example) and want the ability to specify that, while `topic1` uses `AvroSerDe1` by default, I can write to it with `AvroSerDe2` in some cases. Granted, this is an edge case: most of the time I'd only want to use whatever's specified as the topic's defaults. But it'd be nice to have this flexibility just in case.

What if you moved topic -> SerDe assignment out of routing?

```ruby
topic_mapping do
  topic :topic1 do
   # If deserializer is not specified, defaults to Karafka global default (JSON)
    deserializer MyAvro1
   # If serializer is not specified, defaults to topic's deserializer (which, if blank, see above)
    serializer MyAvro1
  end
end
```

mensfeld:
> I like your idea of changing the names ("serializer/deserializer" instead of "parser")

Let's do this!

>  I'm still not sure about specifying topic serializers in this section where we're defining routes.

Your argument is (again) more than valid. I agree. This idea is now rejected ;)

> That might be true in the case of consumers (though even there, not necessarily) but it seems a poor fit for responders.

:+1: 

> This should only be used when serializing messages to the "topic1" topic. But a responder may write to (many) other topics. It would be best to tie serialization to each topic. Is that what you had in mind here? If so consider revising the inline comment as something like:

I didn't but that is more than a fair point. And moving it out of the routes makes a lot of sense. But in that case, I would suggest something else that should cover all the cases:

```ruby
topic_mapping do
  topic :topic1 do
   # If deserializer is not specified, defaults to Karafka global default (JSON)
    deserializer MyAvro1
  end
end
```

And now for the responders serialization:

```ruby
class ExampleResponder < ApplicationResponder
  topic :users_notified, serializer: MySuperCustomXmlSerializer
  topic :users_whatever # default here
  topic :json_users, serializer: MySuperJsonSerializer 

  def respond(user)
    respond_to :users_notified, user
    respond_to :users_whatever, user
    respond_to :json_users, user
  end
end

ExampleResponder.call(User.last)
```

And we would have one deserializer that would be a karafka config option.

benissimo:
> > This should only be used when serializing messages to the "topic1" topic. But a responder may write to (many) other topics. It would be best to tie serialization to each topic. Is that what you had in mind here? If so consider revising the inline comment as something like:
> 
> I didn't but that is more than a fair point. And moving it out of the routes makes a lot of sense. But in that case, I would suggest something else that should cover all the cases:
> 
> ```ruby
> topic_mapping do
>   topic :topic1 do
>    # If deserializer is not specified, defaults to Karafka global default (JSON)
>     deserializer MyAvro1
>   end
> end
> ```
> 
> And now for the responders serialization:
> 
> ```ruby
> class ExampleResponder < ApplicationResponder
>   topic :users_notified, serializer: MySuperCustomXmlSerializer
>   topic :users_whatever # default here
>   topic :json_users, serializer: MySuperJsonSerializer 
> 
>   def respond(user)
>     respond_to :users_notified, user
>     respond_to :users_whatever, user
>     respond_to :json_users, user
>   end
> end
> 
> ExampleResponder.call(User.last)
> ```
> 
> And we would have one deserializer that would be a karafka config option.

This works, but suppose I have multiple responders that write to a topic `:error`. I could specify

```ruby
Responder1 < BaseResponder
  ...
 topic :error, serializer: XMLSerializer
  ...
end

Responder 2 < BaseResponder
  ...
  topic :error, serializer: XMLSerializer
  ...
end
```

but it'd be nice to avoid that repetition by instead specifying the serialization format for `error` topic in the topic mapping section:

```ruby
topic_mapping do
  topic :error do
    serializer: XMLSerializer
```

and only overriding that for special cases:

```
Responder3 < BaseResponder
  topic :error, serializer: Responder3CustomXMLSerializer # overrides topic_mapping assignment, only for this responder
```

benissimo:
> Ok, so we would have to introduce a whole new "entity" in the system, that would allow to describe such cases. Can we postpone that to `1.4` and go for now with the per responder case that seems to cover all the cases apart from introducing duplication? That way we can incorporate it into `1.3` quite fast and think about the extended design for `1.4`.

Sure, so punt on `topic_mapping` for now and just rewrite the responder serializer setup to use:

```ruby
class ExampleResponder < ApplicationResponder
  topic :users_notified, serializer: MySuperCustomXmlSerializer
  topic :users_whatever # default here
  topic :json_users, serializer: MySuperJsonSerializer 
```

I can do that part now, as a rewrite of this pull request.

I'll avoid introducing any changes to the routing code (`consumer_groups.draw ...`) as that's more involved. This way my commit will be backwards-compatible, simply offering a new way to specify serializers for responders.

mensfeld:
> This way my commit will be backwards-compatible, simply offering a new way to specify serializers for responders.

Yes. We can do this and I will take the effort to change `parser` to deserializers for 1.3.

Awesome!

benissimo:
Implemented PR feedback and rebased onto `1.3-wip`

mensfeld:
Great. Will take a look once done with #464 :) Thank you!

mensfeld:
Applied small remarks but overall great work! Thank you!

benissimo:
Thanks @mensfeld!

Issue #463, opened 2019-01-28
Split parsers into serializers / deserializers


mensfeld:
As @benissimo pointed out here: https://github.com/karafka/karafka/pull/461 parsers violate SRP and don't provide enough flexibility as the outgoing messages don't have to match the incoming type.

We should split it into two, replacing the `parser` with `deserializer` and leaving the serializer as it is, only to inject it by default into presenters. The rest will be updated by @benissimo and we can sync on that later.

We will also have to update docs.

PR #464, opened 2019-01-28
splitted responsibilities

mensfeld:
close https://github.com/karafka/karafka/issues/463

cc: @benissimo

note: does not provide way to overwrite the default serializer in the responder as this will be provided in a different PR.


lib/karafka/base_responder.rb
```diff
@@ -82,16 +82,18 @@ class << self
       # @param topic_name [Symbol, String] name of topic to which we want to respond
       # @param options [Hash] hash with optional configuration details
       def topic(topic_name, options = {})
+        options[:serializer] ||= Karafka::App.config.serializer
+        options[:registered] = true
```
benissimo: This makes the `options.merge(registered: true)` below redundant.

benissimo: @mensfeld ^ this is my only feedback.

mensfeld: True! Thanks!

lib/karafka/params/params.rb
```diff
@@ -40,30 +40,28 @@ class Params < Hash
...
-          self['parser'].parse(payload)
+      # @param payload [String] Raw data that we want to deserialize using consumer deserializer
+      # @return [Object] deserialized data
+      def deserialize(payload)
+        Karafka.monitor.instrument('params.params.deserialize', caller: self) do
+          self['deserializer'].call(payload)
         end
-      rescue ::Karafka::Errors::ParserError => e
-        Karafka.monitor.instrument('params.params.parse.error', caller: self, error: e)
+      rescue ::StandardError => e
```
nijikon: I'm not sure about this, can't we catch a more specific error here?

mensfeld: The question is, should require users to raise internal karafka error upon their deserializer execution or should we just allow it to raise whatever it wants and handle that. I would go with the second as it is less problematic for the end user, plus anyhow we don't have the "raw" data injection anymore from at leats `1.1`, so we reraise after the instrumentation.

nijikon: Sounds good

PR #466, opened 2019-01-30
deps bump

mensfeld:
Bumps mostly dry libs to keep them up to date within Karafka ecosystem.


mensfeld:
Note on Coditsu: all the new Rubocop remarks are fixed in 1.3.

PR #494, opened 2019-04-24
sidekiq backend issue 29

mensfeld:
Fixes: https://github.com/karafka/sidekiq-backend/issues/29 by injecting the proper naming from which we remap in case it is missing, so the remapping remaps to a proper worker name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@
 - #475 - Disallow subscribing to the same topic with multiple consumers
 - #485 - Setting shutdown_timeout to nil kills the app without waiting for anything
 - #487 - Make listeners as instances
+- #29 (Sidekiq backend) - Consumer class names must have the word "Consumer" in it in order to work
 
 ## 1.2.11
 - [#470](https://github.com/karafka/karafka/issues/470) Karafka not working with dry-configurable 0.8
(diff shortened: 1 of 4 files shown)


lib/karafka/helpers/class_matcher.rb
```diff
@@ -45,6 +45,13 @@ def match
       #   matcher.name #=> Super2Responder
       def name
         inflected = @klass.to_s.split('::').last.to_s
+        # We inject the from into the name just in case it is missing as in a case like that
```
nijikon:
```
We inject the from into the name just in case it is missing as in a situation like that it would just sanitize the name without adding the "to" postfix. It could create cases when we want to build for example a responder to a consumer that does not have the "Consumer" postfix and would do nothing returning the same name. That would be bad as the matching classes shouldn't be matched to themselves.
```

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@
 - #475 - Disallow subscribing to the same topic with multiple consumers
 - #485 - Setting shutdown_timeout to nil kills the app without waiting for anything
 - #487 - Make listeners as instances
+- #29 (Sidekiq backend) - Consumer class names must have the word "Consumer" in it in order to work
```
nijikon: `- #29 - Consumer class names must have the word "Consumer" in it in order to work (Sidekiq backend)`

PR #501, opened 2019-05-07
Zeitwerk setup for the Karafka gem

mensfeld:
This PR does **not** provide the code reloading. It just installs the Zeitwerk and adds it in favor of our old custom loader.


mensfeld:
Outdated.

Issue #483, opened 2019-04-05
Upgrade dry-validation before releasing 1.3


mensfeld:

mensfeld:
ref https://github.com/karafka/waterdrop/pull/92

PR #515, opened 2019-06-10
dry-validations bump to 1.0

mensfeld:
close https://github.com/karafka/karafka/issues/483


lib/karafka/schemas/consumer_group.rb
```diff
@@ -3,7 +3,9 @@
 module Karafka
   module Schemas
     # Schema for single full route (consumer group + topics) validation.
-    ConsumerGroup = Dry::Validation.Schema do
+    class ConsumerGroup < Dry::Validation::Contract
+      config.messages.load_paths << File.join(Karafka .gem_root, 'config', 'errors.yml')
```
tappleby: Is there supposed to be a space here `Karafka .gem_root`? Noticed ResponderUsageTopic also has the same syntax.

mensfeld: Good catch. Checking why the CI didn't catch that...

mensfeld: Dunno. Not catched. Fixed. Thanks!

tappleby: Apparently it’s actually valid syntax 🤷‍♂️

mensfeld: Yeah. Weird. I believe RuboCop should catch that and Coditsu should fail on Circle. Didn't but fixed. Thank you :heart:

Issue #492, opened 2019-04-21
Use Zeitwerk for code reload in development


mensfeld:
Theoretically we should be able to fairly easy hot reload code with that so we wouldn't have to restart things.

PR #521, opened 2019-06-17
Hot code reload

mensfeld:
close https://github.com/karafka/karafka/issues/492

This PR adds option to reload and redraw all the cached mappings and objects to support hot code reload in development.

It should work both with Zeitwerk and Rails as it does not reload framework code, just the invalidates caches so things are recreated.

Limitations:

-  During the redraw, if new topics are present, they won't be included.
- All the multi batch long running operations will be invalidated when the code is reloaded (to be expected but still worth mentioning).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ jobs:
       before_install:
         - gem update --system
         - gem install bundler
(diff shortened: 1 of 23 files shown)


nijikon:
Please add CHANGELOG entry

Issue #370, opened 2018-06-29
Remove singleton and class level global references


mensfeld:
If we're going to support guilds and provide possibility to operate multiple instances of karafka apps (with optional subsets of topics, etc), we need to remove global references.

mensfeld:
part 1: topics aren't anyhow class level limited

mensfeld:
I'm not assigning any milestones as it is an ongoing process.

ka8725:
I almost completed the write the issue https://github.com/karafka/karafka/issues/546 and only then noticed there is already issue for that, this one. But since it has some more explanation I decided to submit it eventually. So it's a duplicate now, but feel free to drop it if you see no reason in it. Thanks!

mensfeld:
@ka8725 thanks! :) I will reply to you there

mensfeld:
I'm closing this one as it is slowly happening. Not sure if all the things are going to be fixed in 2.0 but many already are.

PR #526, opened 2019-07-01
Reloading rails hooks, cleanup, better templates install

mensfeld:
This PR includes:

- big step forward for: https://github.com/karafka/karafka/issues/370 - removal of the majority of singleton classes
- much better templates for the install process including the bootstrap code for the development reload for both Rails and non-Rails apps
- rename of Schemas to Contracts
- Removal of dry-configurable patches as due to using Zeitwerk, they are no longer needed


lib/karafka/cli/install.rb
```diff
@@ -31,12 +43,20 @@ def call
 
         INSTALL_FILES_MAP.each do |source, target|
           target = Karafka.root.join(target)
-          next if File.exist?(target)
 
-          source = Karafka.core_root.join("templates/#{source}")
-          FileUtils.cp_r(source, target)
+          template = File.read Karafka.core_root.join("templates/#{source}")
```
nijikon: `File.read(Karafka.core_root.join("templates/#{source}"))`

lib/karafka/cli/install.rb
```diff
@@ -31,12 +43,20 @@ def call
...
+          render = ::ERB.new(template, trim_mode: '-').result(binding)
+
+          File.open(target, 'w') { |file| file.write render }
```
nijikon: `File.open(target, 'w') { |file| file.write(render) }`

PR #533, opened 2019-07-22
Allow to use headers in the deserializer object

davydovanton:
Hey man!

I played with 1.3 version and found an interesting case for improvement. Now we can use headers. I played with avro and you told me that it's good to put event type and version (for schema-registry) to the headers of kafka message.

And with it I'll be happy to use deserializer like this:

```ruby
module Parsers
  class AvroParser
    REGESTRY_URL = 'http://localhost:8081/'

    attr_reader :avro

    def initialize
      @avro = AvroTurf::Messaging.new(registry_url: REGESTRY_URL)
    end

    def call(content, subject: '', version: 1)
      avro.decode(content, subject: subject, version: version)
    end
  end
end
```

In this case, I can decode the message by subject and version and get decoded payload for in consumer block. Also, it will provide me to use dry-struct for mapping payload to struct object, like this:

```ruby
def call(content, subject: '', version: 1)
  STRUCTS[subject][version].new(
    avro.decode(content, subject: subject, version: version)
  )
end

class ArticlesSnapshotConsumer < ApplicationConsumer
  def consume
    params_batch.each do |message|
      message.payload # => Dry::Struct object
    end
  end
end
```

WDYT about this?

### DIFF
--- a/lib/karafka/params/params.rb
+++ b/lib/karafka/params/params.rb
@@ -55,10 +55,11 @@ module Karafka
       private
 
       # @param payload [String] Raw data that we want to deserialize using consumer deserializer
+      # option headers [Hash] - optional - Kafka message headers
       # @return [Object] deserialized data
       def deserialize(payload)
         Karafka.monitor.instrument('params.params.deserialize', caller: self) do
-          self['deserializer'].call(payload)
+          self['deserializer'].call(payload, self['headers'])
         end
       rescue ::StandardError => e
         Karafka.monitor.instrument('params.params.deserialize.error', caller: self, error: e)
--- a/lib/karafka/serialization/json/deserializer.rb
+++ b/lib/karafka/serialization/json/deserializer.rb
@@ -8,10 +8,11 @@ module Karafka
       # Default Karafka Json deserializer for loading JSON data
       class Deserializer
         # @param content [String] content based on which we want to get our hash
+        # option headers [Hash] - optional - Kafka message headers
         # @return [Hash] hash with deserialized JSON data
         # @example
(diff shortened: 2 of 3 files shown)


lib/karafka/params/params.rb
```diff
@@ -55,10 +55,11 @@ def deserialize!
       private
 
       # @param payload [String] Raw data that we want to deserialize using consumer deserializer
+      # option headers [Hash] - optional - Kafka message headers
       # @return [Object] deserialized data
       def deserialize(payload)
         Karafka.monitor.instrument('params.params.deserialize', caller: self) do
-          self['deserializer'].call(payload)
+          self['deserializer'].call(payload, self['headers'])
```
davydovanton: I played with 
```ruby
def deserialize(payload, headers = {})
  # ...
end
```

but not sure that it's a good idea. WDYT? I can change it

paneq: * `headers` sounds good, but I wonder why not pass the whole message (so that we also have access to `topic` for example). 

Idea based on: https://www.confluent.io/blog/put-several-event-types-kafka-topic/ and the 3 strategies defined at the end of the article, especially: `TopicRecordNameStrategy`.

davydovanton: good for me, @mensfeld WDYT?

mensfeld: I think @paneq has a point here. We could use the whole message. That way someone could use either headers, body or even topic name to build up the messages.

mensfeld: That is a breaking change, so would be good to ship before rc1 ;) (I'll wait for that)

davydovanton:
I'll reopen this PR with Karafka branch

PR #545, opened 2019-09-24
Makes sure the log directory exists when is possible

robertomiranda:
`ensure_dir_exists` it's causing an issue when is performed over `log` and the directory is read-only.


mensfeld:
Thank you! I'll look into that during the weekend!

mensfeld:
Thank you and sorry for the delay!

mensfeld:
Postmerge review - @robertomiranda why did you use `world_writable?` instead of `writable?`

robertomiranda:
@mensfeldI just wanted keep the File method name 😅. Thanks much for merging this PR 🙇

Issue #547, opened 2019-10-31
SSL Client Cert Key Error


arunkarthick:
Hello,

My Kafka cluster uses SSL certificate for authentication. Everything on the Kafka side looks good and its working fine. The default Consumer & Producer scripts are working fine with the client ssl certs. Even its working fine when I try it with ruby-kafka from rails console like below

```
kafka = Kafka.new(["kafka01:9093", "kafka02:9093", "kafka03:9093"], client_id: 'devconsole', ssl_ca_cert: File.read("#{Rails.root.to_s}/config/ssl/ca_cert.pem"), ssl_client_cert: File.read("#{Rails.root.to_s}/config/ssl/client_cert.pem"), ssl_client_cert_key: File.read("#{Rails.root.to_s}/config/ssl/client_key.pem"))
```

```
kafka.deliver_message("Hello, World!", topic: 'test')
=> nil
```

I verified it from a different console to receive this message and I am able to receive this message which was sent through 'ruby-kafka'.

So for so good with 'ruby-kafka' based connection & message delivery.

But when I run the "bundle exec karafka s", I get the following error message

```
Karafka::Errors::InvalidConfigurationError: {:ssl_client_cert_key=>["is not a valid private key\n"]}
  /Users/neo/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/karafka-1.3.0/lib/karafka/routing/builder.rb:44:in `block in draw'
  /Users/neo/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/karafka-1.3.0/lib/karafka/routing/builder.rb:39:in `each'
  /Users/neo/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/karafka-1.3.0/lib/karafka/routing/builder.rb:39:in `draw'
  /Users/neo/example_rails_app/karafka.rb:51:in `<class:KarafkaApp>'
  /Users/neo/example_rails_app/karafka.rb:19:in `<top (required)>'
  /Users/neo/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/zeitwerk-2.1.10/lib/zeitwerk/kernel.rb:23:in `require'
  /Users/neo/.rbenv/versions/2.5.3/lib/ruby/gems/2.5.0/gems/zeitwerk-2.1.10/lib/zeitwerk/kernel.rb:23:in `require'
[...]

PR #551, opened 2019-11-06
add support for DSA keys

nijikon:
Fixes https://github.com/karafka/karafka/issues/547


mensfeld:
Unrelated to this PR, but could you check if this works for Waterdrop 2.0?

PR #574, opened 2020-02-15
dry-configurable upgrade

solnic:
A couple of small tweaks that make it work with dry-configurable `0.11.0` as well as previous versions.


mensfeld:
@solnic can you make sure, that the same will work (if needed) with https://github.com/karafka/waterdrop? Thank you!

mensfeld:
Ok I've checked it. Works. We can merge :rocket:

Issue #582, opened 2020-04-15
Cannot only define seed brokers in consumer groups


matthewtoschi:
I'm trying to configure seed brokers on consumer_groups rather than in the application's setup. As I understand, this is the suggested approach for connecting to multiple kafka clusters for different topics. But leaving it out of the root setup sets the default value of 127.0.0.1.

I assume that one solution is to take one of my seed brokers and just place it in the app setup but I'm wondering if that's necessary and intended. Alternatively, is the intention that consumer group seed brokers are subgroups and I should put all of my seed brokers in the app setup?

--------

## Expected behavior

Seed brokers are not required to be configured at the root level. If all consumer groups configure their own seed brokers, karafka connects to and subscribes from the clusters without issue.

## Actual behavior

Karafka always attempts to connect to the default seed broker, 127.0.0.1, if none are set in the App setup.

## Steps to reproduce the problem

Using example_app, I reconfigured karafka with seed brokers at the consumer_group level. No connection attempt is made for these hosts. Instead there is an error trying to connection to the default seed broker since none are defined in the App setup.
`ERROR -- : Failed to fetch metadata from kafka://127.0.0.1:9092: Connection refused - connect(2) for 127.0.0.1:9092`

```
diff --git a/karafka.rb b/karafka.rb
index fc8d05f..92840bf 100644
--- a/karafka.rb
+++ b/karafka.rb
@@ -30,9 +30,6 @@ APP_LOADER.eager_load
 #   a structure similar to rails config/routes.rb, etc.
 class App < Karafka::App
   setup do |config|
-    # Karafka will auto-discover kafka_hosts based on Zookeeper but we need it 
-    # to run tests without running kafka and zookeeper
-    config.kafka.seed_brokers = [ENV['KAFKA_HOST'] || 'kafka://127.0.0.1:9092']
     config.client_id = 'example_app'
   end
 
@@ -62,6 +59,7 @@ Karafka.monitor.subscribe(Karafka::Instrumentation::ProctitleL
 App.consumer_groups.draw do
   consumer_group :batched_group do
     batch_fetching true
+    seed_brokers = %w( kafka://myhost.com )
 
     topic :xml_data do
       consumer XmlMessagesConsumer
@@ -84,6 +82,8 @@ App.consumer_groups.draw do
   # @note The backend is totally optional, if you disable it, the game will
   # work as well
   consumer_group :async_pong do
+    seed_brokers = %w( kafka://myotherhost.com )
+
     topic :ping do
[...]

PR #585, opened 2020-04-20
#582 - Fix for not being able to define seed_brokers on a per consumer_group basis

mensfeld:
Fixes https://github.com/karafka/karafka/issues/582

Longer description:

The Kafka client settings were not populated down to the client but taken from the "global" settings. This was an unnoticed bug as it affected only the subset of client settings. Since each consumer group is anyhow built based on per consumer group settings OR (if not defined) based on the defaults, this fetches the proper settings directly from the consumer group object, ensuring, that the settings are always the one "closest" to the consumer group.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - #583 - Use Karafka.logger for CLI messages (prikha)
+- #582 - Cannot only define seed brokers in consumer groups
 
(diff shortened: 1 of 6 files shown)


mensfeld:
@nijikon travis is not updating statuses to gh so here's the build link: https://travis-ci.org/github/karafka/karafka/builds/677229177

PR #592, opened 2020-06-24
Fix CHANGELOG.md inaccuracy WRT exp backoff

afhammad:
Support for exponential backoff in pause was not added until 1.3.0


mensfeld:
Thank you!

Issue #598, opened 2020-07-28
Message deserialization: should we allow/simplify metadata access without a deserialization attempt?


rabotyaga:
It is impossible to fetch message metadata (headers, topic, partition, offset) without a deserialization attempt for non-batch consumers using the `params` method.
- for non batch consumers `params` is `params_batch.first`: https://github.com/karafka/karafka/blob/03cc0134ae/lib/karafka/consumers/single_params.rb#L11
- `ParamsBatch#first` calls `deserialize!`: https://github.com/karafka/karafka/blob/03cc0134ae/lib/karafka/params/params_batch.rb#L42
- after deserialization error payload left intact and the message is marked as deserialized: https://github.com/karafka/karafka/blob/03cc0134ae/lib/karafka/params/params.rb#L47

This leads to a probably inconsistent behavior: in case the message could not be properly deserialized, we receive `Karafka::Errors::DeserializationError` on the first attempt to access metadata (all subsequent attempts will finish successfully). Inconsistency is in fact that we tried to access metadata only.

We can work around this by using `params_batch.to_a.first` instead of `params`, then deserialization attempt would be skipped and we would be able to access message metadata.

But wouldn't it be more consistent to allow metadata access without a deserialization attempt?
Consumers could be interested in ability to save/track malformed messages.

--------

## Expected behavior

Probably we should make a deserialization attempt on payload access.

## Actual behavior

Deserialization attempt is made on metadata access.

## Steps to reproduce the problem

- Setup a simple karafka app with an inline consumer
- Insert `binding.pry` into Consumer's `consume` method
- Send `test` (without quotes) using the `kafka-console-producer`
- try to call `params.headers` / `params.topic` / `params.partition` / `params.offset` - you will get `Karafka::Errors::DeserializationError` on the first attempt
  - after each attempt press ^D to continue the consumer loop and reset message deserialization state
- try to call any of the methods above for the `params_batch.to_a.first` - you will be able to use metadata without raising `Karafka::Errors::DeserializationError`

## Your setup details

```
$ [bundle exec] karafka info
Karafka version: 1.3.6
Ruby version: 2.6.5
Ruby-kafka version: 1.1.0
Application client id: platform
Backend: inline
Batch fetching: true
Batch consuming: false
Boot file: /lib/platform_karafka.rb
Environment: development
Kafka seed brokers: ["kafka://localhost:9092"]
```

mensfeld:
Great report! We will fix that in the upcoming week. Thank you!

[...]

PR #599, opened 2020-07-30
Allow metadata access without deserialization attempt

rabotyaga:
Fixes https://github.com/karafka/karafka/issues/598

~~However, it introduces another inconsistency: now `params['payload']` would return the payload without a deserialization attempt. @mensfeld please take a look, I'm not sure if this approach is valid or there is a better way.
If this **is** the way, probably we should override `Karafka::Params::Params#[]` to remove the inconsistency described above?~~

We decided to override `Hash#[]` method to remove any inconsistencies in payload access.
The small benchmark made shows there is no significant performance penalty in such an approach.

### DIFF
--- a/lib/karafka/consumers/single_params.rb
+++ b/lib/karafka/consumers/single_params.rb
@@ -8,7 +8,7 @@ module Karafka
 
       # @return [Karafka::Params::Params] params instance for non batch consumption consumers
       def params
-        params_batch.first
+        params_batch.to_a.first
       end
     end
   end
--- a/lib/karafka/params/params.rb
+++ b/lib/karafka/params/params.rb
@@ -40,6 +40,14 @@ module Karafka
         end
       end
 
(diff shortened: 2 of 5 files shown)


rabotyaga:
I've made a small benchmark for overriding `Hash#[]` method, it seems there is no performance penalty in such approach:
https://gist.github.com/rabotyaga/cf9682e88fdd6cf90b00651c9c3c6702

mensfeld:
@nijikon can we rebase this to use GH actions?

mensfeld:
@rabotyaga LGTM. Feel free to override the `[]` method to align the behavior and we will be good to go. Great work!

rabotyaga:
@mensfeld done

chaadow:
@mensfeld Done #605  and thanks! 👌 yeah I was as surprised as you are!

Issue #602, opened 2020-08-22
1.4 Roadmap


mensfeld:
Params as a hash make it problematic to abstract division in between params metadata and the payload. While we already have payload as a separate entity, the same should happen with metadata. That way we can access and operate on metadata without forcing deserialization of the payload.

I have a POC for `2.0` with this, but due to https://github.com/karafka/karafka/pull/599#issuecomment-678447151 I want to backport it and release `1.4`.

- [x] Rename `Karafka::Params::Metadata` to `Karafka::Params::BatchMetadata` (https://github.com/karafka/karafka/pull/603)
- [x] Separate metadata (including Karafka native metadata) from the root of params (https://github.com/karafka/karafka/pull/604)
- [x] Add missing specs to params metadata (https://github.com/karafka/karafka/pull/604)
- [x] Remove metadata hash dependency
- [x] Remove params dependency on a hash in favour of PORO (https://github.com/karafka/karafka/pull/604)
- [x] Remove batch metadata dependency on a hash  (https://github.com/karafka/karafka/pull/604)
- [x] Remove MultiJson in favour of JSON (ref https://github.com/karafka/karafka/issues/605)
- [x] Release Waterdrop 1.4
- [x] Update example-app and check the touched components (especially manual offset management)
- [x] Update the karafka-testing lib
- [x] Update the sidekiq-backend
- [x] Release `1.4-alpha`
- [x] Release `1.4-rc1`
- [x] Prepare release notes
- [x] Update wiki
- [x] Release `1.4`

PR #604, opened 2020-08-22
Separate Params::Metadata from Params

mensfeld:
Part 2 of https://github.com/karafka/karafka/issues/602

We extract `Karafka::Params` details that are not related to payload but are metadata, so they can be accessed independently, leaving `Params` as a thin line connecting the payload and the metadata.

Overall this PR:

- allows accessing all the metadata without accessing the payload
- removes the hash dependency from all the params elements
- freezes params and underlying elements except for the mutable payload
- provides access to raw payload after serialization
- fixes the bug where a non-deserializable (error) params would be marked as deserialized after first unsuccessful deserialization attempt
- fixes bug where karafka would mutate internal ruby-kafka state
- fixes bug where topic name in metadata would not be mapped using topic mappers
- simplifies the params and params batch API, before `#payload` usage, it won't be deserialized
- removes the `#[]` API from params to prevent from accessing raw data in a different way than `#raw_payload`
- makes the params batch operations consistent as params payload is deserialized only when accessed explicitly


nijikon (approved):
Nice!

rabotyaga:
@mensfeld Very nice changes / fixes!

Should we add backward compatibility for metadata attributes? I mean, you can use `params.headers` for a big variety of versions up to 1.3.x, and then in 1.4 everything suddenly blows up.
Something like https://github.com/karafka/karafka/pull/609

Issue #605, opened 2020-08-23
`MultiJson.load` does not accept `nil` as a param compared to JSON ruby library


chaadow:
@mensfeld I'm using MultiJson `1.15.0` along with `json` gem version `2.3.1` I believe these are the latest versions

But I believe this is not an adapter issue. from my screenshot above i went to inspect the line in the multijson gem:

https://github.com/intridea/multi_json/blob/master/lib/multi_json/adapter.rb#L20

and they raise an error if the string is blank, which is the case when you pass `nil` however if you deserialize directly with the JSON library there is no error

_Originally posted by @chaadow in https://github.com/karafka/karafka/pull/599#issuecomment-678687284_

mensfeld:
I think that the inconsistency comes from the RFC update.

This bug actually aligns with my desire to remove multi-json as a "not needed" dependency, thus nice timing :)

ref: https://tools.ietf.org/html/rfc7159

Also: https://github.com/intridea/multi_json/pull/152

I will be removing MultiJson and anyone that needs it can easily add it by using custom serializer :) this should:

1. fix the inconsistency
2. fix your bug
3. remove the additional dependency

mensfeld:
@chaadow as a temp bypass just use a custom serializer that references JSON directly.

mensfeld:
@chaadow could you test it? https://github.com/karafka/karafka/pull/606

mensfeld:
fixed here: https://github.com/karafka/karafka/pull/606

chaadow:
> @chaadow could you test it? #606

Added my review in #606

PR #606, opened 2020-08-24
Remove multi-json

mensfeld:
fixes https://github.com/karafka/karafka/issues/605


nijikon (approved):

davydovanton (approved):
👍

lib/karafka/serialization/json/deserializer.rb
```diff
@@ -17,8 +17,8 @@ class Deserializer
         #   }
         #   Deserializer.call(params) #=> { 'a' => 1 }
         def call(params)
-          ::MultiJson.load(params.raw_payload)
-        rescue ::MultiJson::ParseError => e
+          ::JSON.parse(params.raw_payload)
```
chaadow, line 20: Actually JSON.parse `nil` raises nil compared to JSON.load

mensfeld: :facepalm:

mensfeld: @nijikon can we add a regression for this? I changed it just because of damn Rubocop :man_shrugging: in the last commit

mensfeld: One thing that I can say for sure after all of this, is that I'm by no means a JSON expert @chaadow

chaadow: @mensfeld Neither am I 😂 all good 👌

PR #607, opened 2020-08-24
JSON fix

mensfeld:
Fix for handling nil with JSON parse that is a valid Kafka case for log compaction.


mensfeld:
@chaadow  not merging without your approval this time ;)

chaadow:
@mensfeld you got my approval 👌

PR #609, opened 2020-08-26
Add backward compatibility for params metadata attributes

rabotyaga:
Related to: https://github.com/karafka/karafka/pull/604

### DIFF
--- a/lib/karafka/params/params.rb
+++ b/lib/karafka/params/params.rb
@@ -7,7 +7,11 @@ module Karafka
     # using deserializer until we execute our logic. That way we can operate with
     # heavy-deserialization data without slowing down the whole application.
     class Params
+      extend Forwardable
+
       attr_reader :raw_payload, :metadata
+
+      def_delegators :metadata, *Metadata.members
 
       # @param raw_payload [Object] incoming payload before deserialization
       # @param metadata [Karafka::Params::Metadata] message metadata object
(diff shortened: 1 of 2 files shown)


mensfeld:
I don't think we should. That's why it's a breaking changes release `1.4` instead of `1.3.x`. This is backported by me from 2.0 but I'm open for discussion.

rabotyaga:
I have no strong opinion here, if this was exactly the plan then it's perfectly fine for me.
Just had thought at first, that this breaking change is too much breaking for incrementing only the minor number.

mensfeld:
Actually after giving it a second look and thought I think we can merge this. You're arguement is strong and due to 2.0 I feel we can add it to 1.4 and remove in 2.0.

I really appreciate your help 😊

mensfeld:
Would you mind reopening? I would love to merge it to rc2

rabotyaga:
Thanks!

mensfeld:
🙏

rabotyaga:
Should I add specs or wrap with some kind of deprected warnings?

mensfeld:
Specs are always welcomed :)

I don't think we need to add deprec. 2.0 bump should be enough with extensive upgrade notes that I always release.

One thing worth considering is using a direct struct keys reference for methods forwarding instead of referencing manually. That way if we would add some sort of metadata in 1.4.x we would not have to remember to update it anywhere else expect in the metadata struct and it's builder.

mensfeld (approved):

PR #630, opened 2020-11-14
Add support for :assignment_strategy for consumers

Azdaroth:


lib/karafka/contracts/consumer_group.rb
```diff
@@ -93,6 +94,10 @@ class ConsumerGroup < Dry::Validation::Contract
         end
       end
 
+      rule(:assignment_strategy) do
+        key.failure(:does_not_respond_to_call) if value && !value.respond_to?(:call)
```
mensfeld, line 98: does nil or false respond to call? if not, then this is redundant and since as default we use roundrobin instance that responds to call, this seems redundant

lib/karafka/setup/config.rb
```diff
@@ -89,6 +89,9 @@ class Config
         #   are stored for further processing. Note, that each item in the queue represents a
         #   response from a single broker
         setting :fetcher_max_queue_size, 10
+        # option assignment_strategy [Object] a strategy determining the assignment of
+        #   partitions to the consumers.
+        setting :assignment_strategy, Kafka::RoundRobinAssignmentStrategy.new
```
mensfeld, line 94: could we wrap it with Karafka simpledelegator? Lets say in a `Karafka::AssignmentStrategies::RundRobin`? I would prefer not to directly show the Kafka internals if possible

lib/karafka/contracts/consumer_group.rb
Azdaroth: Good call, makes sense

lib/karafka/setup/config.rb
Azdaroth: Sure

spec/lib/karafka/contracts/consumer_group_spec.rb
```diff
@@ -48,6 +48,7 @@
       max_bytes_per_partition: 1_048_576,
       offset_retention_time: 1000,
       fetcher_max_queue_size: 100,
+      assignment_strategy: Kafka::RoundRobinAssignmentStrategy.new,
```
mensfeld, line 51: shouldn't it be:
```ruby
Karafka::AssignmentStrategies::RoundRobin.new
```

Azdaroth: it should, fixed ;)

mensfeld:
@nijikon could you look why the hell our specs are not running?

mensfeld (approved):

mensfeld:
@Azdaroth would you mind looking into this? it sems that round robin have changed in ruby-kafka and the spec is failing: https://github.com/karafka/karafka/pull/704

Ref: https://my.diffend.io/gems/ruby-kafka/1.3.0/1.4.0#d2h-437175

I would appreciate help :pray:  thank you

Azdaroth:
hi @mensfeld , I'll check it out later this week

PR #660, opened 2021-03-14
Change `rails?` check to  `railties` gem

ojab:
`rails` is just a meta-gem with a bunch of dependencies, all the Rails stuff can be installed without it and depends on `railties`.


ojab:
Okay, making changes/PR from the GH web-interface is not a good thing.

mensfeld (approved):

PR #662, opened 2021-03-15
Remove pidfile only if it was created & clarify comment

ojab:


mensfeld:
Thank you. FYI pidfile support will be completely removed from 2.0.

mensfeld (approved):

PR #676, opened 2021-04-16
Remove rake dependency

bilby91:
We are trying to use Karafka with an application that uses an old version of rake. After analyzing the commits, I realized that rake is no longer used by this project but the dependency is still present in the gemspec. 

This pull request removes the dependency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Remove Ruby 2.5 support and update minimum Ruby requirement to 2.6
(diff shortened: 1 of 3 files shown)


mensfeld (approved):

mensfeld:
Thank you. I will release it next week.

mensfeld:
@bilby91 thank you once again for your work. This update has been released: https://github.com/karafka/karafka/pull/678

mensfeld:
https://my.diffend.io/gems/karafka/1.4.3/1.4.4

PR #712, opened 2021-09-15
Support exactly once delivery and transactional messaging

ojab:
karafka config doesn't allow it, so additional `WaterDrop` configuration is needed and it needs to be done carefully, because karafka will overwrite it to `false` with it's own config on initialization.

### DIFF
--- a/lib/karafka/connection/api_adapter.rb
+++ b/lib/karafka/connection/api_adapter.rb
@@ -12,6 +12,10 @@ module Karafka
     #   do nothing. So we don't have to worry about injecting our internal settings
     #   into the client and breaking stuff
     module ApiAdapter
+      # Settings that are not accepted by `ruby-kafka` for consumer, but allowed for producer
+      PRODUCER_ONLY_SETTINGS = %i[idempotent transactional transactional_timeout].freeze
+      private_constant :PRODUCER_ONLY_SETTINGS
+
       class << self
         # Builds all the configuration settings for Kafka.new method
         # @param consumer_group [Karafka::Routing::ConsumerGroup] consumer group details
@@ -41,6 +45,7 @@ module Karafka
           end
 
           settings_hash = sanitize(settings)
+          PRODUCER_ONLY_SETTINGS.each { |x| settings_hash.delete(x) }
 
           # Normalization for the way Kafka::Client accepts arguments from  0.5.3
           [settings_hash.delete(:seed_brokers), settings_hash]
--- a/lib/karafka/contracts/consumer_group.rb
+++ b/lib/karafka/contracts/consumer_group.rb
@@ -67,6 +67,10 @@ module Karafka
         # It's not with other encryptions as it has some more rules
         optional(:sasl_scram_mechanism)
           .maybe(:str?, included_in?: SASL_SCRAM_MECHANISMS)
+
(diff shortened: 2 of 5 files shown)


mensfeld:
Hey @ojab, thank you for your contribution. Do you want my help in finishing this or are you good?

ojab:
Gonna finish it probably today, gotta deal with contract specs.

ojab:
Ok, added some specs, not sure if anything else should be added. Ready for review, I guess.

mensfeld:
Thank you. I will give it a look in the upcoming days.

ojab:
Added spec that checks `Karafka::Setup::Configurators::WaterDrop`, wouldn't hurt.

mensfeld:
@ojab did you try to initialize WaterDrop yourself with your own configuration? Did it not work?

ojab:
Yep, this works and used right now, but it's not ideal: kafka config should be duplicated because exact example from wiki doesn't work:
```
$ bundle exec ruby k.rb 
/home/ojab/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/waterdrop-1.4.2/lib/water_drop/config.rb:158:in `validate!': {:kafka=>{:seed_brokers=>["must be an array"]}} (WaterDrop::Errors::InvalidConfiguration)
	from /home/ojab/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/waterdrop-1.4.2/lib/water_drop/config.rb:144:in `block in setup'
```
at least seed broker and who knows what else should be duplicated and I'm not even sure if configuring waterdrop after `App.boot!` is [thread]safe.

So IMHO it's better to allow configuration on karafka level.

mensfeld:
@ojab I see your point now. In 2.0 that is handled as you can completely overwrite the producer.

> at least seed broker and who knows what else should be duplicated and I'm not even sure if configuring waterdrop after App.boot! is [thread]safe.

It is. Boot is bootstraping things but does not run anything on it's own. Server does but as long as the configuration is after boot, but not somewhere inside the code, nothing will happen.

Maybe we should just expose whole scope of settings for the producer? Maybe we should just expose the configuration API for this case instead of configuration settings, WDYT?

ojab:
Yep, IMHO exposing plain {rd,ruby-}kafka configuration is the proper option here, but not sure if anything should be done for 1.4, probably it would be quite intrusive.
I'm even fine with closing this PR, because right now it works via separate waterdrop configuration.

mensfeld:
@ojab if you don't mind I will rework it to support custom waterdrop insertion.

Despite 2.0 I still plan to support 1.4 for a longer period of time, thus this change would be valuable to me and to other users.

mensfeld:
@ojab but please reopen - I plan to rework it based on your work :D

ojab:
oh, no problem

mensfeld:
Ok, update from me: closing. Focusing all my efforts on finalizing 2.0. Please reopen if needed.

rammpeter:
Are there any further plans for supporting transactional producer in Karafka?

mensfeld:
Yes: https://github.com/karafka/waterdrop/issues/243 

This is scheduled for execution right after the Karafka Web UI that I am finishing

Zrzut ekranu z 2023-01-05 19-10-41

PR #746, opened 2021-12-30
Ruby 3.1

mensfeld:


mensfeld:
rails does not work with ruby 3.1 for now and we use activejob specs. Will wait for now.

PR #749, opened 2022-01-01
remove irb dep

mensfeld:
irb is part of stdlib, thus it is not needed to include it in the gemfile


nijikon (approved):

st0012:
Not suggesting to add it back, but Rails 7.1 will have IRB as [its dependency](https://github.com/rails/rails/blob/main/railties/railties.gemspec#L47). The reason is that recent versions of IRB have received significant amount of enhancements and bug fixes between 2.7, 3.2, and the upcoming 3.3. By installing it as a gem, users in older Ruby versions can also get a decent experience in the console instead of getting stuck with an outdated one. We're also working on [extension APIs](https://github.com/ruby/irb/pull/624) to make it a better platform for gems to built their console with it.

That being said, I'm aware that karafka-core still supports Ruby 2.6 while IRB is already considering dropping 2.7's support (due to YARP/Prism). So I just want to mention that there is and will be benefit of having IRB as a dependency despite it being a default gem 🙂

mensfeld:
@st0012 thanks for the suggestion. Correct me if I'm wrong but can't user just define it himself if he wants?

```ruby
gem 'karafka'
gem 'irb'
``` 

that way, once that want to have it slim will have it, but those that want newer will just use a newer one?

PR #755, opened 2022-01-03
optimize context switching in the jobs queue

mensfeld:
This PR changes how we skip listener thread execution and pass it to workers. Previous approach would `Thread.pass` effectively running upon each scheduler switch just to pass it. The new approach blocks with a queues (one per group) until it is unblocked when needed. This effectively reduces context switching from 200k switches to listener to just over 200. (99% reduction).

Since we always wait on all the workers work to be done both on polling and shutdown, we can do it that way.

inspired by @ioquatix <3

### DIFF
--- a/bin/stress
+++ b/bin/stress
@@ -9,5 +9,5 @@ set -e
 while :
 do
   reset
-  bundle exec bin/integrations
+  bundle exec bin/integrations $1
 done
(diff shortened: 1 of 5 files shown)


nijikon:
This is a very very nice change! 💜

PR #765, opened 2022-01-17
Restore code-reload support in development mode

mensfeld:
It will be slow in this case: https://github.com/rails/rails/issues/44183 but is irrelevant to the PR

### DIFF
--- a/spec/integrations/consumption/without_persistence_enabled.rb
+++ b/spec/integrations/consumption/without_persistence_enabled.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+# Karafka should not use same consumer instance when consumer_persistence is set to false
+# Each batch of data should be consumed with new instance
+
+setup_karafka do |config|
+  config.consumer_persistence = false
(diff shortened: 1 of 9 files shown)


nijikon:
Love this! 💜

spec/integrations/consumption/without_persistence_enabled.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+end
+
+elements.each { |data| produce(DataCollector.topic, data) }
+
+start_karafka_and_wait_until do
+  DataCollector.data[0].size >= 100
+end
+
+assert_equal 100, DataCollector.data[0].size
```
chaadow, line 33: I'm wondering if you can add a test ( or edit this one) to check if all object ids are distinct. my idea would be to convert `DataCollector.data[0]` to a ruby `Set` and  check that its size is 100 @mensfeld

mensfeld: https://github.com/karafka/karafka/pull/771

PR #782, opened 2022-02-09
Config hash must respect symbols as a strings

bragovo:
```
it { expect(group.kafka['bootstrap.servers']).to eq(topic.kafka[:'bootstrap.servers']) }
```
This test passed but because both are nil 😭 You need to add the next line for it
```
it { expect(group.kafka['bootstrap.servers']).not_to be_nil }
```

I think we need to stringify keys for this hash.


mensfeld:
closing as superseeded by https://github.com/karafka/karafka/pull/783 thank you for rising this issue!

Issue #803, opened 2022-03-18
Support 2.0.0 Rails integration without ActiveJob


mensfeld:
ref https://github.com/tshpachenko-ccycloud/karafka-non-aj/blob/main/Gemfile

PR #805, opened 2022-03-18
Rails support without ActiveJob queue adapter usage

mensfeld:
ref https://github.com/tshpachenko-ccycloud/karafka-non-aj/blob/main/Gemfile

close https://github.com/karafka/karafka/issues/803

This PR ensures that we can use Karafka with Rails without ActiveJob. It also ensures that rspec/rails as well as other frameworks do not detect dummy ActiveJob namespace without a reason when ActiveJob is not loaded.


mensfeld:
Confirmed to work. Merging.

PR #833, opened 2022-04-18
backoffspecs

mensfeld:
This PR adds missing specs for a cases of backoff handling

### DIFF
--- a/spec/integrations/consumption/non_critical_error_exponential_backoff.rb
+++ b/spec/integrations/consumption/non_critical_error_exponential_backoff.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+# When we have non-critical error happening couple times and we use exponential backoff, Karafka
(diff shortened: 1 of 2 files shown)


spec/integrations/consumption/non_critical_error_exponential_backoff.rb
```diff
@@ -0,0 +1,61 @@
...
+
+start_karafka_and_wait_until do
+  DataCollector.data[0].size >= 10
+end
+
+# Backoff time before next exception occurrence (not before resume). Because of that and the fact
+# that we run this in parallel, we add some extra time to compensate.
+BACKOFF_RANGES = [
+  0..0.5,
+  0..1,
```
nijikon, line 39: Why do we need 0..1 three times?

mensfeld: Because this is how the backoff algorithm works with smaller ranges.

PR #848, opened 2022-05-05
Allow time based pausing from the consumers

mensfeld:
This PR allows us to pause for an explicitly provided amount of time.

It introduces to the consumer `#pause` and `#resume` methods + adds both unit and integration specs.

This can be useful for when manually controlling the pausing (advanced usage).


nijikon:
Great change, code is clean.

PR #903, opened 2022-07-18
Remove dry-validation in favour of a home-brew validation layer

mensfeld:
This PR removes our dependency on `dry-validation` without any changes to the end users.


lib/karafka/contracts/config.rb
```diff
@@ -9,71 +9,90 @@ module Contracts
...
+        config.error_messages = YAML.safe_load(
+          File.read(
+            File.join(Karafka.gem_root, 'config', 'errors.yml')
+          )
+        ).fetch('en').fetch('validations').fetch('config')
+      end
+
+      # License validity happens in the licenser. Here we do only the simple consistency checks
+      nested(:license) do
+        required(:token) { |token| [true, false].include?(token) || token.is_a?(String) }
```
nijikon, line 22: We could simplify this by using `val` in the block. Based on the argument's name, we already know what we are operating on.

nijikon (approved):

nijikon (approved):

Issue #979, opened 2022-08-18
Expose failure reason in the processing result


mensfeld:
As a part of transporting the error notion forward via coordinator, we need to track the error reason when it occurs.

PR #980, opened 2022-08-18
track failure cause in processing

mensfeld:
This PR introduces processing failure cause tracking for things like DLQ.

close https://github.com/karafka/karafka/issues/979

### DIFF
--- a/lib/karafka/processing/result.rb
+++ b/lib/karafka/processing/result.rb
@@ -6,8 +6,11 @@ module Karafka
     # It allows to indicate if given thing moved from success to a failure or the other way around
     # Useful for tracking consumption state
     class Result
+      attr_reader :cause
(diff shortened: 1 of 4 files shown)


lib/karafka/processing/result.rb
```diff
@@ -18,11 +21,16 @@ def success?
       # Marks state as successful
       def success!
         @success = true
+        # We set cause to false so the previous error that occurred does not leak when error is
```
nijikon, line 24: I think that the comment should be above cause

nijikon (approved):

lib/karafka/processing/result.rb
mensfeld: yeah, the order was fixed.

Issue #988, opened 2022-08-21
Maunally paused partition is not processed forward after a re-claim


mensfeld:
This does not affect LRJ. It does affect regular processing.

If partition is lost and later re-claim and was paused for a really long time, the processing is not picked up if there is a re-claim.

PR #989, opened 2022-08-21
Fix a indefinite manual pause after rebalance

mensfeld:
close https://github.com/karafka/karafka/issues/988

This spec also adds an integration case for this case.

We do not run this for LRJ as LRJ will automatically resume after finishing (via resume when success or short pause when failed)

### DIFF
--- a/lib/karafka/pro/base_consumer.rb
+++ b/lib/karafka/pro/base_consumer.rb
@@ -44,6 +44,29 @@ module Karafka
         end
       end
 
+      # Trigger method for running on partition revocation.
+      #
+      # @private
+      def on_revoked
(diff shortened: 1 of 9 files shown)


lib/karafka/pro/base_consumer.rb
```diff
@@ -74,6 +97,8 @@ def on_after_consume_regular(first_message, last_message)
           resume
         else
           # If processing failed, we need to pause
+          # For long running job this will overwrite the default never-ending pause and will cause
+          # the processing th keep going after the error backoff
```
nijikon, line 101: `the processing to keep going after the error backoff`

nijikon:
LGTM 🚀

Issue #1002, opened 2022-08-29
Multiple LRJ topics fetched concurrently block ability for LRJ to kick in


mensfeld:
Running multiple LRJs with concurrency set to one will cause the consumer to be kicked out of the consumer group because pausing will not happen. 

We can mitigate this by running the job phases in separation, similar to how we run shutdowns. That way all the pre work will happen prior to LRJ kicking in.

PR #1003, opened 2022-08-30
Prevent multiple LRJ from blocking polling on pausing

mensfeld:
close https://github.com/karafka/karafka/issues/1002

This PR ensures that even when having multiple LRJ topics with multiple partitions in the same consumer group, they are not starved because of low concurrency. This could cause polling to be blocked.

No public API changes.


nijikon:
Looks good 💜

PR #1037, opened 2022-09-21
Optimize error handling edge cases + improve routing validation and seek offset notion

mensfeld:
In this PR:

- Marking as consumed at the end of processing should be async not to trigger any rebalance errors as there will be no error handling anyhow, and after a potential unsuccessful marking it will poll proper data (if rebalance)
- `coordinator_load_in_progress` error for forced committing offsets should retry as it is a temporary issue that should transition to a different state (either successful or post balance error)
- `coordinator_load_in_progress` should also have a retry and backoff in the listener loop
- use backoff strategy for any errors related to polling
- do not reset client upon `rebalance_in_progress` and just backoff
- introduce new integration specs to cover the `manual_offset_management` error handling without any marking on multiple batches
- extract seek offset to coordinator, so it can be accessible from all the consumers when needed (especially from VPs)


lib/karafka/connection/client.rb
```diff
@@ -329,25 +336,27 @@ def poll(timeout)
         #
         # If we would retry here, the client reset would become transparent and we would not have
         # a chance to take any actions
+        early_return = false
```
nijikon, line 339: Can we rename this to `early_exit`?

nijikon (approved):
LGTM 🚀

lib/karafka/connection/client.rb
nijikon: Sounds good

PR #1044, opened 2022-09-24
Improve error notifications on key polling errors

mensfeld:
I gave it a second, though, and I think we need to make one change to the reporting of network / critical logic errors here:

while we can and should still retry them, on those errors, we should early notify them before retries. The reasoning is that network and other criticals indicate severe system problems, and max poll interval exceeding means severe business logic problems (slow processing).

This code does not change how the retry engine works but it provides early important errors notifications.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Karafka framework changelog
 
+## Unreleased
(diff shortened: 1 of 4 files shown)


nijikon (commented):
LGTM 🚀

PR #1050, opened 2022-10-03
cooperative-sticky rebalance strategy support + offset commit improvements

mensfeld:
This PR:
- improves the default librdkafka rebalance callback to flush the offsets upon revocation. This can help reduce number of messages that are re-processed.
- provides support for cooperative-sticky assignment strategy + integration specs about it
- User instrumentation publishes internal rebalance errors instead of using rdkafka logger

This PR does not change any APIs.


nijikon (approved):
LGTM 🚀

PR #1057, opened 2022-10-11
rdkafka v0.13.0 support

mensfeld:
This PR makes karafka require 0.13 or more. Since does not bring any breaking changes it should not be a problem to update.


nijikon (approved):

mensfeld:
will fail for now until waterdrop is released but I run it locally against patched waterdrop and lgtm.

mensfeld:
closing for now. Once 0.13 is available we can then fetch upstream and do it without keeping branches like this open anymore

PR #1067, opened 2022-10-14
v1.4.14

mensfeld:


mensfeld:
Failing specs fail because of the changes in ruby-kafka that are beyond maintenance PR.

PR #1096, opened 2022-10-20
Point to correct location of Commercial license

soulcutter:
I'm not sure the mechanics of changing licenses. It may be that it would be preferable to rename the commercial license file instead.

Can you believe that people read the license? ;)


mensfeld:
Thank you. I read it few times after "borrowing" from Sidekiq :laughing:  (which Mike said is ok btw)

PR #1100, opened 2022-10-24
Reorganize features management

mensfeld:
This PR:

- Introduces a concept of routing features. We have some things like manual offset management, ActiveJob, a few pro features, and DLQ (in the future), plus potentially others that are, from the routing perspective, pretty much self-sufficient, and they should not have to be embedded into the code directly as their routing settings + validations can act in isolation.
- Removes the `manual_offset_management` options from config as it is now extracted into a feature
- Moves `ActiveJob` routing extensions into a separate feature as well
- close https://github.com/karafka/karafka/issues/1099

The increased complexity in `BaseConsumer` is expected as it will be a subject to different PR to extract feature and feature combinations flows for consumers into a separate set of handlers.


lib/karafka/routing/features/active_job/config.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Karafka
+  module Routing
+    module Features
+      class ActiveJob < Base
+        # Config for ActiveJob usage
+        Config = Struct.new(:active, keyword_init: true) do
```
nijikon, line 8: I would like us to be consistent with the formatting.
```ruby
Config = Struct.new(
  :active,
  keyword_init: true
) { alias_method :active?, :active }
```

lib/karafka/routing/features/manual_offset_management/config.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Karafka
+  module Routing
+    module Features
+      class ManualOffsetManagement < Base
+        # Config for manual offset management feature
+        Config = Struct.new(:active, keyword_init: true) do
```
nijikon, line 8: Same formatting comment here
```ruby
Config = Struct.new(
  :active,
  keyword_init: true
) { alias_method :active?, :active }
```

nijikon (approved):
LGTM 🚀

PR #1102, opened 2022-10-25
Do not load Railtie for Rails integration when KARAFKA_RAILTIE_LOAD is set to false

mensfeld:
see https://github.com/karafka/example-apps/pull/190

When Karafka is bundled within internal gem that holds configuration and common components, Railtie may kick in prior to configuration and requirement of this gem, messing up the load order.


nijikon (approved):
LGTM 🚀

mensfeld:
No longer needed.

PR #1103, opened 2022-10-25
Migrate strategy selection for regular (non pro) consumption combinations

mensfeld:
This PR moves the current processing flow from consumers into strategies.

Having separate strategies allows us to remove edge cases, `ifs` and other things. Each feature combination supported will have it's own strategy.

In this PR we move only current non PRO strategies out of the consumer.


lib/karafka/base_consumer.rb
```diff
@@ -32,6 +41,13 @@ def on_before_enqueue; end
     def on_before_consume
       messages.metadata.processed_at = Time.now
       messages.metadata.freeze
+    rescue StandardError => e
```
nijikon, line 44: Why don't we have `handle_before_consume` here?

mensfeld: because we never needed it in any scenario. I will add it if there is a scenario that would need it.

PR #1122, opened 2022-11-07
Datadog::LoggerListener instrumentation

mensfeld:
Followup to the great work of @bruno-b-martins

https://github.com/karafka/karafka/commit/4732f43b02f4db426d404b6c87877cdbafe3647c


mensfeld:
@bruno-b-martins  would you mind taking a look? I've extracted the client out of the instrumentation so I can inject a dummy into it and added an integration spec to make sure that the integration works as expected.

Alongside did small cleanups in the specs support vendors to align the conventions.

mensfeld:
> BTW, I'd probably just change the name of spec/integrations/instrumentation/data_dog_logger_flow_spec.rb to .../datadog_logger_flow_spec.rb (data_dog => datadog).

True. Will fix that.

lib/karafka/instrumentation/vendors/datadog/logger_listener.rb
```diff
@@ -0,0 +1,153 @@
...
+            error
+            fatal
+          ].freeze
+
+          private_constant :USED_LOG_LEVELS
+
+          # @param block [Proc] configuration block
+          def initialize(&block)
+            configure
+            setup(&block) if block
```
bruno-b-martins, line 34: Hey @mensfeld 👋 

Can you please explain me why do we need both the `configure` and the `setup(&block)` that calls `configure(&block)`?

My guess is that you're covering two use cases:
1. when someone instantiates this class and only afterwards changes the `client` or `config`.
2. when someone instantiates this class and passes a block to set up `client` or `config`.

Is that right?

mensfeld:
> Don't know if it makes sense to add something like that to the docs

Yes will be added.

lib/karafka/instrumentation/vendors/datadog/logger_listener.rb
mensfeld: Yes exactly. This is to align the API with other components.

mensfeld:
Remarks were pushed. Waiting on the CI and will merge.

Issue #1151, opened 2022-11-13
Provide ability to quiet Karafka


mensfeld:
We would like to be able to tell karafka not to poll new work but to keep processing what it does and then stop processing without shutting down.

This API needs to include the API for `Embedding`.

PR #1153, opened 2022-11-15
#1151 Quiet

mensfeld:
close https://github.com/karafka/karafka/issues/1151

This PR adds the `TSTP` signal support, so we silence karafka when it needs to keep running without accepting new jobs.


nijikon (approved):
LGTM

Issue #1148, opened 2022-11-10
Prevent multiple close operations on clients happening at the same time


mensfeld:
We should not run `close` at the same time on multiple consumers because it causes things to hang:

```ruby
MUTEX = Mutex.new

1000.times.map do
  other = Thread.new do
    consumer = setup_rdkafka_consumer
    #MUTEX.synchronize do
      consumer.close
    #end
  end
end.each(&:join)
```

vs

```ruby
1000.times.map do
  other = Thread.new do
    consumer = setup_rdkafka_consumer
    MUTEX.synchronize do
      consumer.close
    end
  end
end.each(&:join)
```

mensfeld:
Now I'm not sure if this is an issue. Will have to investigate further...

PR #1156, opened 2022-11-16
Optimize shutdown

mensfeld:
This PR optimizes shutdown of a single consumer group with subscription groups in which the processing performance varies.

close https://github.com/karafka/karafka/issues/1154
close https://github.com/karafka/karafka/issues/1148


nijikon:
LGTM

PR #1162, opened 2022-11-19
Remove unused constant

mensfeld:
This error class is no longer in use


nijikon (approved):

spec/support/data_collector.rb
```diff
@@ -7,6 +7,11 @@
 class DataCollector
   include Singleton
 
+  # Mutext we use to ensure we don't have multi-threaded issues when collecting data
```
nijikon, line 10: Typo

Issue #1173, opened 2022-11-23
Support static group membership with multiple subscription groups


mensfeld:
without it:

```ruby
rdkafka: [thrd:main]: Fatal error: Broker: Static consumer fenced by other consumer with same group.instance.id: Fatal consumer error: Broker: Static consumer fenced by other consumer with same group.instance.id
rdkafka: [thrd:main]: Fatal error: Broker: Static consumer fenced by other consumer with same group.instance.id: Fatal consumer error: Broker: Static consumer fenced by other consumer with same group.instance.id
rdkafka: [thrd:main]: Fatal error: Broker: Static consumer fenced by other consumer with same group.instance.id: Fatal consumer error: Broker: Static consumer fenced by other consumer with same group.instance.id
rdkafka: [thrd:main]: Fatal error: Broker: Static consumer fenced by other consumer with same group.instance.id: Fatal consumer error: Broker: Static consumer fenced by other consumer with same group.instance.id
rdkafka: [thrd:main]: Fatal error: Broker: Static consumer fenced by other consumer with same group.instance.id: Fatal consumer error: Broker: Static consumer fenced by other consumer with same group.instance.id
rdkafka: [thrd:main]: Fatal error: Broker: Static consumer fenced by other consumer with same group.instance.id: Fatal consumer error: Broker: Static consumer fenced by other consumer with same group.instance.id
rdkafka: [thrd:main]: Fatal error: Broker: Static consumer fenced by other consumer with same group.instance.id: Fatal consumer error: Broker: Static consumer fenced by other consumer with same group.instance.id
rdkafka: [thrd:main]: Fatal error: Broker: Static consumer fenced by other consumer with same group.instance.id: Fatal consumer error: Broker: Static consumer fenced by other consumer with same group.instance.id
```

PR #1174, opened 2022-11-23
Support of static group membership with multiple subscription groups

mensfeld:
This PR provides support for working with static group memberships where there are multiple subscription groups.

close https://github.com/karafka/karafka/issues/1173


nijikon (approved):
LGTM 🚀

PR #1193, opened 2022-11-28
Few small fixes and remarks

mensfeld:
This PR:
- allows for usage of symbols in subscription groups routing definitions to align the API allowed for topics and consumer groups
- removes allowance of nil in the block definition of sub group
- fix invalid subscription groups count displayed in the `karafka info` command


nijikon (approved):
LGTM

PR #1195, opened 2022-11-28
Move consumption to strategies

mensfeld:
This PR moves the consume handler to strategies so we can customize this for LRJ and others when needed.


nijikon (approved):
LGTM

PR #1198, opened 2022-11-29
Allow running particular named subscription groups and topics similar to consumer groups

mensfeld:
close https://github.com/karafka/karafka/issues/1192


nijikon (approved):
LGTM

Issue #1155, opened 2022-11-16
Duplicated logs  in development environment


Michanio:
## Expected behavior

Karafka should not extend the logger when the logger output is set to `$stdout`.
When the output is different than `$stdout` Karafka should extend the logger but the new logger that writes to `$stdout` should use the same formatter as its defined for the original one.  

## Actual behavior

When logger's log device is set $stdout we can see duplicated log messages on the output of development environment.
Moreover the logs are formatted differently when custom formatter is defined.

## Steps to reproduce the problem

Use the changes applied [here](https://github.com/karafka/example-apps/pull/193) and run the karafka server.
You should see sth like:

```
@@@                                             @@@@@  @@@
@@@                                            @@@     @@@
@@@  @@@    @@@@@@@@@   @@@ @@@   @@@@@@@@@  @@@@@@@@  @@@  @@@@   @@@@@@@@@
@@@@@@     @@@    @@@   @@@@@    @@@    @@@    @@@     @@@@@@@    @@@    @@@
@@@@@@@    @@@    @@@   @@@     @@@@    @@@    @@@     @@@@@@@    @@@    @@@
@@@  @@@@  @@@@@@@@@@   @@@      @@@@@@@@@@    @@@     @@@  @@@@   @@@@@@@@@@


[CustomFormatter] [2022-11-16T10:11:51.133363] INFO --
@@@                                             @@@@@  @@@
@@@                                            @@@     @@@
@@@  @@@    @@@@@@@@@   @@@ @@@   @@@@@@@@@  @@@@@@@@  @@@  @@@@   @@@@@@@@@
@@@@@@     @@@    @@@   @@@@@    @@@    @@@    @@@     @@@@@@@    @@@    @@@
@@@@@@@    @@@    @@@   @@@     @@@@    @@@    @@@     @@@@@@@    @@@    @@@
@@@  @@@@  @@@@@@@@@@   @@@      @@@@@@@@@@    @@@     @@@  @@@@   @@@@@@@@@@


Upgrade to Karafka Pro for more features and support: https://karafka.io
[CustomFormatter] [2022-11-16T10:11:51.133396] INFO -- Upgrade to Karafka Pro for more features and support: https://karafka.io
Running in ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [arm64-darwin21]
[...]

PR #1200, opened 2022-11-30
Fix duplicated logs in development environment

mensfeld:
close https://github.com/karafka/karafka/issues/1155

alongside improved poro/pristine detection and added:
```ruby
ENV['KARAFKA_CLI'] = 'true'
```

to make sure all the Railtie stuff is initialized.


nijikon (approved):
LGTM

PR #1214, opened 2022-12-09
Introduce terminated state

mensfeld:
This PR introduces `terminated` state to differentiate between the stopped (no consumption, no connections via listeners, etc) and the last step, which is termination (exit).

The distinction is needed because we need to be able to instrument on stopped in the web UI and via listeners. `terminated` will run right after exit.


nijikon:
LGTM

PR #1216, opened 2022-12-10
use shorter ids

mensfeld:
This PR aligns the id format with waterdrop and makes it shorter

aligns with waterdrop: https://github.com/karafka/waterdrop/pull/280

note: We don't have to use `hex(12)` for really unique ids (like Sidekiq) because we do not assign ids per each job and rest of the objects like threads, consumers, executors are long living.


nijikon (approved):
LGTM

PR #1258, opened 2023-01-09
add checksum verification specs

mensfeld:
This PR specs out the integration verification script we are going to recommend for pro users.


mensfeld:
Irrelvant, there is much better way.

PR #1273, opened 2023-01-19
better dlq keys strategy

mensfeld:
This PR changes the DLQ dispatch key to make the distribution better in case of multi-partition single DLQ topic.


nijikon (approved):
LGTM 🚀

PR #1309, opened 2023-02-13
More reliable admin operations

mensfeld:
There are cases where the create topic requests are queued in Kafka and not yet finished while the promise is materialized. This means we can create a topic and expect it to work but "it's not there" yet.


mensfeld:
Nvmd. That is not the issue.

Issue #1325, opened 2023-02-20
Topic name validation issue


sastromskis:
PR #1300 introduced an updated regex for topic naming validation which throws an error in case the topic name contains both dot and underscore. 

We are using Debezium MySQL CDC connector on confluent cloud, which by default creates topics having exactly the forbidden naming convention:

`server.database_name.table_name`

Would it be possible to make this validation optional?

mensfeld:
> Would it be possible to make this validation optional?

Yes.

sastromskis:
Thank you!

mensfeld:
@nijikon @CandyFet My suggestion for this would be to validate potential collisions in the scope of Karafka routing: that is only failing if there are topics that would collide. For example:

`server.database_name.table_name` and `server.database.name.table_name` within the routing (any consumer groups). WDYT?

I think this would be less restrictive but in a context of Karafka would still ensure correctness. I don't think we should be that strict to limit overall scope of operations.

mensfeld:
Ok others have same problem. Example: `xyz-web.development.data_flows.avro`

mensfeld:
Merged. I will release it this week. Just need to finish one or two more things.

sastromskis:
@mensfeld Thank you!

mensfeld:
@sastromskis exactly. We want to promote good practices but sometimes you already have what you have ;) just set it to false and it won't cause you any problems anymore!

PR #1326, opened 2023-02-20
#1325 Lift the strict topic naming convention limitation

mensfeld:
This PR relaxes the topic naming validation. Many people argued that the newly introduced naming validation needs to be more relaxed. 

While I understand the need to ensure users follow good patterns, many users using external sink tools encounter problems because of this validation. It was reported several times. I'm lifting this validation in favour of usability with other tooling that generates topics names out of the box and to make sure Karafka can be easily integrated into already existing Kafka systems.

close https://github.com/karafka/karafka/issues/1325

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - [Fix] Karafka monitor is prematurely cached (#1314)
 - [Improvement] Make `::Karafka::Instrumentation::Notifications::EVENTS` list public for anyone wanting to re-bind those into a different notification bus.
 - [Improvement] Set `fetch.message.max.bytes` for `Karafka::Admin` to `5MB` to make sure that all data is fetched correctly for Web UI under heavy load (many consumers).
+- [Fix] Remove inconsistent topic naming validation check. It was causing problems for many production grade systems already using certain internal naming conventions.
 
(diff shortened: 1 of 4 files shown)


mensfeld:
@CandyFet happy to get your opinion on that one.

nijikon (approved):

nijikon:
@mensfeld maybe we could leave this in, but you would have to enable strict topic validation, which would be off by default.

mensfeld:
@nijikon we would have to introduce a flag. The question would be: should we introduce it on a root level (`config.strict_topic_validation`) or introduce a new namespace (preferred by me): `config.routing.strict_topics_validations.strict = false`

nijikon:
@mensfeld whatever you think is better, if it is only related to `routing`, then `config.routing.strict_topics_validations` is fine

CandyFet:
Hi guys, sorry for late response. I think that config flag is a great idea! I see that a lot of people struggling with this new validations but people who uses karafka in production and have topic name validations with ops team might have benefits with tracking correct namespacing in topic names

mensfeld:
Ok. I'm closing this one in favour of a strict flag that will be disabled by default.

PR #1345, opened 2023-02-25
1128 - Removed Pro::BaseConsumer in favor of BaseConsumer.

### LINKED ISSUE #1128: Remove Karafka::Pro::Base consumer
Because now we have a way to inject strategies on the regular consumer to make pro seamless, no need to inherit. With a major we can remove this Pro consumer that now only wraps the base.

federicomoretti:
Closes #1128.

With the big refactor in 2.0 we no longer need a dedicated class for the Base Consumer in Pro.
CHANGELOG has been updated with instructions on how to migrate as part of 2.1


Gemfile.lock
```diff
@@ -67,6 +67,7 @@ GEM
     zeitwerk (2.6.7)
 
 PLATFORMS
+  arm64-darwin-21
```
federicomoretti, line 70: Thoughts on adding m1?

federicomoretti:
@mensfeld can I get your thoughts here? I don't know what you had in mind.

Gemfile.lock
mensfeld: yeah it's ok. Its just a lockfile so for missing platforms it will just materialize itself if missing.

mensfeld:
@federicomoretti it's good but I would still keep the `Karafka::Pro::BaseConsumer` as an empty shell that inherits from Base. The reason is: I want to have backwards compatibility for first few pro users that used to use it. I will reach out to them and once I confirm all moved away from it, we will be able to safely remove it.

federicomoretti:
@mensfeld can we release this as part of 2.1? I'm happy to write the CHANGELOG entry. No need to merge this PR soon.

mensfeld:
@federicomoretti I would merge it without removal of the Pro consumer but just as an alias and create a separate card to remove it totally with 2.1. I don't follow semantic versioning with Karafka (though I probably should) as I usually keep the second version (2.x) as an indicator of a bigger change.

mensfeld:
@federicomoretti it's marked as draft, is it ready? (no rush, just asking)

federicomoretti:
@mensfeld I'm sorry I never saw your comment. Do you want me to target 2.1 and remove the consumer completely? Or else I could make the Pro Consumer to just be an alias of the BaseConsumer, although I'm not sure how that's better than what we have today (inheritance with no change on behavior).

mensfeld:
We can easily get rid of it as 2.1 is a major and breaking changes are ok ;) I want to get rid of this and few other legacy things.

mensfeld:
Retargeting so I can fix rspec. No need to do anything else on your side. Merging :)

PR #1423, opened 2023-04-16
Stability log format and early break

mensfeld:
This PR:

- Improves logger listener reporting for jobs by including the topic partition number
- Normalizes the format of partition reporting in logger listener to always use topic/partition
- Improves stability by doing early break on invalid topic during shutdown not to end up with a really long (or never ending) shutdown
- fixes flaky spec

### DIFF
--- a/lib/karafka/connection/client.rb
+++ b/lib/karafka/connection/client.rb
@@ -369,6 +369,8 @@ module Karafka
       rescue ::Rdkafka::RdkafkaError => e
         early_report = false
 
+        retryable = time_poll.attempts <= MAX_POLL_RETRIES && time_poll.retryable?
+
         # There are retryable issues on which we want to report fast as they are source of
         # problems and can mean some bigger system instabilities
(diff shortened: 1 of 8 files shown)


lib/karafka/connection/client.rb
```diff
@@ -389,9 +391,10 @@ def poll(timeout)
           return nil if @subscription_group.kafka[:'allow.auto.create.topics']
 
           early_report = true
-        end
 
-        retryable = time_poll.attempts <= MAX_POLL_RETRIES && time_poll.retryable?
+          # No sense in retrying when no topic/partition and we're no longe running
```
servercrunch, line 395: longe -> longer

nijikon, line 395: `longe` -> `longer`

mensfeld: Thanks fixed

nijikon (approved):

Issue #1389, opened 2023-03-29
Stand up one-off consumers for investigation


aemadrid:
## Is your feature request related to a problem? Please describe.

Sometimes we need to investigate issues and we need to query a topic to find an answer.

## Describe the solution you'd like

I would like to stand up a ephemeral consumer to query the topic like:

```ruby
options = {
  from: Time.new(2023, 2, 1),
  to: Time.new92023, 3, 1),
  limit: 500
}
consumer = Karafka.consume('topic_name', options)
messages = consumer.select{|message| message.dig('a', 'b') == 123 }
```

The consumer would allow us to run through all messages in that timeframe so we could deal with only the data that we need.

## Describe alternatives you've considered

We could stand up a proper consumer and write code to process the messages that we need the way we want to. the issue would be that it would involved a full deployment cycle for an investigation that is truly one-off.

mensfeld:
rel: https://github.com/appsignal/rdkafka-ruby/pull/229

mensfeld:
Few things we need to tackle:

1. We need to support both abilities to run on one partition and on many. This means that we should not provide any limit (except batch size) and instead yield batches until we reach EOF for all partitions or until someone breaks
2. When using timestamp lookup, it should be one but it should apply to all partitions that someone wants to provide.
3. multi-topic should not be supported

My suggested API would look something like that:

Case 1: offset based

```ruby
Karafka::Admin.stream_topic(:topic_name, partitions: { 0 => 123,  1 => 456}, max_messages: 100) do |messages|

end
```

Case 2: timestamp based:

```ruby
# If no partitions, defaults to all
Karafka::Admin.stream_topic(:topic_name, partitions: [1, 2, 3], timestamp: 2.days.ago, max_messages: 100) do |messages|

end
```

wdyt?

mensfeld:
done and in master, i will create a separate card for the timestamp based lookup support

PR #1426, opened 2023-04-18
Iterating API for one-off lookups

mensfeld:
This PR provides offset based iteration API. It does **not** provide time seeking as this will be provided as a separate initiative.

ref https://github.com/karafka/karafka/issues/1389


nijikon (approved):
Great work on this 💪

PR #1434, opened 2023-04-28
963 - Add support for Current Attributes to ActiveJob's adapter.

federicomoretti:
WIP


lib/karafka/active_job/consumer.rb
```diff
@@ -24,21 +24,26 @@ def consume
...
+          # We publish both to make it consistent with `consumer.x` events
+          Karafka.monitor.instrument('active_job.consume', payload)
+          Karafka.monitor.instrument('active_job.consumed', payload) do
+            ::ActiveJob::Base.execute(job)
+          end
         end
       end
+
+      # @param job_message [Karafka::Messages::Message] message with active job
+      def with_deserialized_job(job_message)
```
federicomoretti, line 41: Ideally, if we had a middleware system like Rack or Sidekiq, we could hook the CurrentAttributes code into that. 
For the current approach, I created two methods in the Dispatcher and Consumer so that we have a way to hook CurrentAttributes into the job's serialization/deserialization process.

spec/lib/karafka/pro/active_job/current_attributes_spec.rb
```diff
@@ -0,0 +1,112 @@
+# frozen_string_literal: true
```
federicomoretti, line 1: I duplicated the spec for Pro because the Pro Dispatcher overrides the dispatch method and handles serialization itself. I wanted to make sure both Dispathers and Consumers (Pro and non-pro) used CurrentAttribues when enabled.

mensfeld:
retargeting so I can try to change it to initilizable modules. Amazing work! Just amazing!

PR #1440, opened 2023-05-11
963 - Add support for Current Attributes to ActiveJob's adapter. (#1434)

mensfeld:


mensfeld:
@federicomoretti quick question:

can I register couple Current objects? Similar to Sidekiq? Because based on my understanding it will not work:

```ruby
  #   # in your initializer
  #   require "sidekiq/middleware/current_attributes"
  #   Sidekiq::CurrentAttributes.persist("Myapp::Current")
  #   # or multiple current attributes
  #   Sidekiq::CurrentAttributes.persist(["Myapp::Current", "Myapp::OtherCurrent"])
```

mensfeld:
@federicomoretti I:
- refactored it a bit
- added integration specs for OSS and Pro + concurrency checks
- added support for multiple current contexts similar to sidekiq. 

I would appreciate your review :pray:

federicomoretti:
@mensfeld this is great! I actually had the same question when I was looking at Sidekiq's approach, why would they not support multiple Current Attribute classes. But I see that they actually added this 4 days ago 😃 
I like the extra checks and specs, this looks solid. I can update this PR to match: https://github.com/karafka/wiki/pull/24

PR #1460, opened 2023-05-28
A test case for message ordering during rebalance with synchronous commits

Aerdayne:
Submitting this as a PR since I'm not sure if it's an actual issue or not. It might feel a bit all over the place for which I apologise.

I've been tinkering with a potential scenario where a specific message out of a batch hangs the consumer for whatever reason - this can be either due to a misconfigured timeout on the downstream service, or something worse.

Here's an additional caveat: the consumer deals with ordered messages. Imagine a chain of events such as `create`, `update`, `update` all interacting with the same entity. Reprocessing the first `update` could bring non-deterministic results, and yes, the second `update` would be most likely also reprocessed, but business logic implemented by the handler might not be ready for it. Ideally the body of a handler should be idempotent, retryable and cover as many corner cases as possible, but it's not always feasible.

Whether already processed messages will get reprocessed relies completely on the mode of interacting with the broker. I guess the `LRJ` feature would help with this since it pauses during consumption, **BUT** it seems to me that it's intended for cases where the nature of work that consumer will perform is known in advance. Increasing the poll interval would also help, but just like `LRJ`, it implies that all messages processed by a particular consumer are expected to take long to process.

However, my case is aimed at providing a guarantee to generic ordered consumers in case of exceptional failures, where during the normal system health most of the work doesn't take long to execute. Note that retrying the problematic message itself is totally fine and expected, and obviously unavoidable.
In this case, seems like there are two options: either setting `max_messages` to 1, so that every `poll` fetches only a single message, or committing the offset synchronously for each message with `#mark_as_consumed!`. It would be ideal not to limit the amount of messages fetched by a `poll`, but still provide a guarantee that messages will not be reprocessed out of order, so `#mark_as_consumed!` would be a better choice.

Please let me know if my reasoning is sound, or if there's another intended way of implementing this.

I wrote a spec that makes sure that with `#mark_as_consumed!` being used in a consumer that cares about order, no already processed messages should be retried. **I added a twist on top of it in the form of a rebalance, and this is where the potential issue arises. Please refer to the inline comments.**

Since the integration spec suite is quite extensive and can be read as some form of documentation, would you think that introducing a spec for showcasing that without `MOM` or auxiliary `#mark_as_consumed!` calls, it's possible for ordered messages to be **re**processed out of order? And another spec like the one in this PR, which makes sure that already processed messages are NOT reprocessed when `#mark_as_consumed!` is used? These cases are already somewhat covered by other specs, but there's no explicit test where the issue of reordering is brought to the forefront (please correct me if there is).

If you feel like these specs would be a good addition, I can adjust this and introduce the other one if you want.


spec/integrations/rebalancing/message_order_with_sync_commit.rb
```diff
@@ -0,0 +1,162 @@
...
+absent_return_value = DT[:mark_as_consumed_return_values].find do |key, _|
+  key == DT[:contested_message].first.payload['key']
+end
+assert_equal absent_return_value, nil
+
+# ... it will raise Rdkafka::RdkafkaError instead
+assert_equal DT[:errors].count, 1
+consumption_error = DT[:errors].first
+assert_equal consumption_error.code, :unknown_member_id
+assert_equal consumption_error.class, Rdkafka::RdkafkaError
```
Aerdayne, line 162: Here's where the potential problem is: when the first consumer 'wakes up' and tries to synchronously commit the message after it has already lost the partition, #mark_as_consumed! raises Rdkafka::RdkafkaError with `unknown_member_id` code, which is not rescued by `Connection#internal_store_offset` and `false` is not returned. Is this intended?

spec/integrations/rebalancing/message_order_with_sync_commit.rb
```diff
@@ -0,0 +1,162 @@
...
+        sleep(0.1) until DT.data.key?(:second_closed)
+      end
+
+      DT.data[:consecutive_messages] << [message.payload['key'], message.partition, message.offset]
+
+      # #mark_as_consumed! will raise on the contesed message since by that time
+      # the partition will be revoked
+      return_value = mark_as_consumed!(message)
+      DT[:mark_as_consumed_return_values] << [message.payload, return_value]
+      return if return_value == false
```
Aerdayne, line 38: All of the assertions in this spec showcase the behaviour that is expected, except for this construct here (assertions are commented on at the bottom of the spec).

spec/integrations/rebalancing/message_order_with_sync_commit.rb
```diff
@@ -0,0 +1,162 @@
...
+      # Wait until we receive a batch with more than 3 messages, where the current one follows
+      # at least 2 other messages in order to avoid false positives and give the second consumer
+      # a chance to reprocess the whole batch.
+      condition_met = ((messages.metadata.first_offset + 2)..(messages.metadata.last_offset - 1)).cover?(message.offset)
+
+      # Wait until some amount of messages has been already processed to spice things up
+      if condition_met && DT.data[:contested_message].empty? && DT.data[:consecutive_messages].count >= 60
+        DT[:contested_message] << message
+        sleep(0.1) until DT.data.key?(:second_closed)
+      end
```
Aerdayne, line 30: Simulate a failure where the message take too long to process for some reason. This will wait for the 2nd consumer to process it (meaning that it will wait until `poll` interval is exceeded and rebalance happens) in order to simulate the worst case scenario.

spec/integrations/rebalancing/message_order_with_sync_commit.rb
mensfeld: If `unknown_member_id` is not cought, it's definitely a bug. Let me try to create an isolated spec for it...

About your other things: yeah all over the place so I need to wrap by head around it :sweat_smile:

mensfeld:
ref https://github.com/karafka/karafka/issues/1461

spec/integrations/rebalancing/message_order_with_sync_commit.rb
mensfeld: @Aerdayne it's a bug. I was able to reduce the spec needed and will shortly commit the fix. Then we can get back to the original discussion.

mensfeld: @Aerdayne fix: https://github.com/karafka/karafka/pull/1462

mensfeld:
Ok, aside from the error fixed in 2.1.3 lets bisect your comment :sweat_smile: 

> Submitting this as a PR since I'm not sure if it's an actual issue or not. It might feel a bit all over the place for which I apologise.

No need to apologize. Your contributions are extremely valuable.

> I've been tinkering with a potential scenario where a specific message out of a batch hangs the consumer for whatever reason - this can be either due to a misconfigured timeout on the downstream service, or something worse.

You are not the only one. It's really important however to define clearly the difference in between long processing and unintentional hanging (which you did).

> Whether already processed messages will get reprocessed relies completely on the mode of interacting with the broker. I guess the LRJ feature would help with this since it pauses during consumption, BUT it seems to me that it's intended for cases where the nature of work that consumer will perform is known in advance. Increasing the poll interval would also help, but just like LRJ, it implies that all messages processed by a particular consumer are expected to take long to process.

LRJ is intended for cases where you know you will exceed max.poll. In general I recommend setting up batch processing in such a way (aside from LRJ), to not to exceed 50% of max.poll ms. The rest should act as a buffer. Which means by defaults as far as I remember 2.5 minute per batch.

It is also worth keeping in mind, that Karafka aims to minimize number of concurrent connections to Kafka via subscription groups and in case of an oversaturation of work, 2.5 minute per batch may be more than expected collectively :sweat_smile: .

> However, my case is aimed at providing a guarantee to generic ordered consumers in case of exceptional failures, where during the normal system health most of the work doesn't take long to execute. Note that retrying the problematic message itself is totally fine and expected, and obviously unavoidable.

In cases like this you may want to look into Karafka Kubernetes supervisor, which can be tuned to kill the process while before the max.poll.interval.ms or session timeout are exceeded. Ofc. it's not a perfect solution but it's one you may want to have in a robust system where you are concerned about things you have mentioned.

> In this case, seems like there are two options: either setting max_messages to 1, so that every poll fetches only a single message, or committing the offset synchronously for each message with #mark_as_consumed!

I presented you with a third option. Setting up supervision to ensure that work that exceeds expected times (considered a flaw of the system) causes a kill via supervisor.

> It would be ideal not to limit the amount of messages fetched by a poll, but still provide a guarantee that messages will not be reprocessed out of order, so #mark_as_consumed! would be a better choice.

Your notion of "out of order" is not exactly right. The moment you exceed max.poll.interval and other "hard" timeouts, the current process (hanging) is considered dead. If it is doing any work, it's a design and setup flaw, not Karafka per se. However I do see your point here :) Karafka makes sure to commit the offsets on rebalances as long as work is done, providing really good warranties within the proper timeouts. This + cooperative.sticky and static group memberships should allow for absolute minimum of reprocessings (again: within timeouts).

I've been exploring ways to "auto-enable" LRJ when needed for scenarios like yours (prior to overexceeding, auto switch to LRJ), however this creates other problems.

> Please let me know if my reasoning is sound, or if there's another intended way of implementing this.
It makes absolute sense :)

> I wrote a spec that makes sure that with #mark_as_consumed! being used in a consumer that cares about order, no already processed messages should be retried. I added a twist on top of it in the form of a rebalance, and this is where the potential issue arises. Please refer to the inline comments.

This spec is amazing! Thank you :pray: I will have to slightly align it but aside from that works great!

> Since the integration spec suite is quite extensive and can be read as some form of documentation, would you think that introducing a spec for showcasing that without MOM or auxiliary #mark_as_consumed! calls, it's possible for ordered messages to be reprocessed out of order?

But it is not. Your case is not far from a scenario where I would spawn a separate thread doing some updates on DB after the partition is lost. It's up to the end user to obey the time limitations imposed by Kafka and Karafka. Karafka also provides the Web-UI that can be used to inspect those errors easily and mitigate that. There always needs to be a limit to processing time defined, otherwise it would not be possible to trigger recovery rebalances from Kafka perspective.

> And another spec like the one in this PR, which makes sure that already processed messages are NOT reprocessed when #mark_as_consumed! is used? These cases are already somewhat covered by other specs, but there's no explicit test where the issue of reordering is brought to the forefront (please correct me if there is).

There are many specs that ensure, that marking messages as consumed behaves as expected within and outside the time limitations but I am always happy to accept more :D If you could not find what you were looking for, happy to accept a PR illustrating your use-case.

> If you feel like these specs would be a good addition, I can adjust this and introduce the other one if you want.
Totally!

That said, you are getting out of order because you are going beyond constraints that are defined exactly to manage such cases. It's not different from having a session timeout on a DB, a rolled-back transaction and a second process waiting for the lock and a restart of the transaction (again) effectively breaking the locking order (assuming hard order requirement). It is again a design flaw and not the DB flaw. If your operations are such sensitive, you could use the last offset mark as a per entity version schema track not to reply old (applied) operations again via filtering inside of queries (as long as DB) or use the offset in other ways to ensure the consistency. That said, it's still a bypass to your underlying problem which is a processing flow that spans from 0 to infinity. Every system needs to have constraints within which it operates and some ways of ensuring those are met (supervision, alerting), I don't think there is a way around that. And if your processing is really unpredictable, then LRJ was meant exactly for that (not an exception but expectation).

Aerdayne:
> In cases like this you may want to look into Karafka Kubernetes supervisor, which can be tuned to kill the process while before the max.poll.interval.ms or session timeout are exceeded. Ofc. it's not a perfect solution but it's one you may want to have in a robust system where you are concerned about things you have mentioned.

Yep, I've already looked into the k8s instrumentation that Karafka has and it would definitely help with such cases.

> The moment you exceed max.poll.interval and other "hard" timeouts, the current process (hanging) is considered dead. If it is doing any work, it's a design and setup flaw, not Karafka per se.

Totally agree, it's impossible to implement a generic safeguard on the infrastructural framework layer. Ultimately when such guarantees are needed, the handler's business logic should be aware of it regardless.

>  This + cooperative.sticky and static group memberships should allow for absolute minimum of reprocessings (again: within timeouts).
👍 

> But it is not.

Not sure I follow. If you remove `#mark_as_consumed!` in this spec (and do some other obvious tweaks), then some messages will be reprocessed after the ones following them. Here's a slice from a `contested_messages` array (substituting keys with consumer denominator) after a test run:

```
 [:first, 1, 19], [:first, 0, 20], [:first, 0, 21], [:second, 0, 20], [:second, 0, 21], [:second, 0, 22]
```

Message with offset 20 is clearly reprocessed after message 21.
I'm not implying this is Karafka's fault since clearly the handler code violates all constraints and time limitations. My point is that with a synchronous `#mark_as_consumed!` it would be impossible, even though it will incur a performance cost. The other spec that I mentioned would be basically that - a slightly tweaked submitted spec with no sync commit. And just to clarify, since you've fixed the `unknown_member_id` problem, the 'issue' aspect of this PR is resolved 🙂

> That said, you are getting out of order because you are going beyond constraints that are defined exactly to manage such cases. It's not different from having a session timeout on a DB, a rolled-back transaction and a second process waiting for the lock and a restart of the transaction (again) effectively breaking the locking order (assuming hard order requirement). It is again a design flaw and not the DB flaw. If your operations are such sensitive, you could use the last offset mark as a per entity version schema track not to reply old (applied) operations again via filtering inside of queries (as long as DB) or use the offset in other ways to ensure the consistency. That said, it's still a bypass to your underlying problem which is a processing flow that spans from 0 to infinity. Every system needs to have constraints within which it operates and some ways of ensuring those are met (supervision, alerting), I don't think there is a way around that. And if your processing is really unpredictable, then LRJ was meant exactly for that (not an exception but expectation).

You are right about all of this. Introducing a DB-level domain-aware fencing mechanism which would store offsets right next to the side effects that are being emitted by the code would be the most robust solution. But alas, not always possible due to various reasons, human factor being one of them 😕. Sync commit would be a good middle ground since it's generic enough to just plug into a base consumer, while providing a guarantee as well.

Thanks for such a detailed response and solid advice. I'll add the other test case in the following days.

mensfeld:
> Not sure I follow. If you remove #mark_as_consumed! in this spec (and do some other obvious tweaks), then some messages will be reprocessed after the ones following them. Here's a slice from a contested_messages array (substituting keys with consumer denominator) after a test run:

Yes. I was referring to a case, where you anyhow end up with at least one duplicate:

```
 [:first, 1, 19], [:first, 0, 20], [:first, 0, 21], [:second, 0, 21], [:second, 0, 22]
```

sorry for not being clear enough. I was also referring to the fact, that within time boundaries defined when working with Kafka, Karafka commits all the offsets during rebalance on revocation, prior to further work assignment, making it (in theory) impossible to assign work in such a way, that would case even a single message duplication processing.

mensfeld:
Retargeting to apply some remarks (not worth discussing over and over again) and will merge. Thank you!

mensfeld (approved):

Issue #1461, opened 2023-05-28
`mark_as_consumed!` raises an error instead of `false` on `unknown_member_id`


mensfeld:
Integration spec to come.

Issue #1459, opened 2023-05-28
Spec failing randomly under heavy load


mensfeld:
This spec:

```
consumption/strategies/dlq/lrj_mom/non_recoverable_moving_forward_marking_spec.rb
```

has a tendency to fail under heavy load during stress tests.

PR #1462, opened 2023-05-28
Make sure to handle unknown_member_id when dealing with revoked partition on marking

mensfeld:
close https://github.com/karafka/karafka/issues/1461

### DIFF
--- a/lib/karafka/connection/client.rb
+++ b/lib/karafka/connection/client.rb
@@ -305,6 +305,8 @@ module Karafka
         case e.code
         when :assignment_lost
           return false
+        when :unknown_member_id
+          return false
         when :no_offset
           return true
         when :coordinator_load_in_progress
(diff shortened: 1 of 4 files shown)


mensfeld:
Note to self: failing spec here is the one we already know is not fully stable: https://github.com/karafka/karafka/issues/1459

mensfeld:
@Aerdayne I don't have a way to assign you to this PR, but your :eye: would be appreciated :pray:

lib/karafka/connection/client.rb
```diff
@@ -305,6 +305,8 @@ def internal_commit_offsets(async: true)
         case e.code
         when :assignment_lost
           return false
+        when :unknown_member_id
+          return false
```
Aerdayne, line 309: I guess `internal_store_offset` can not cause an error with this code since it does not contact the broker directly, and that's why you're not rescuing it there? I'm just double checking since `assignment_lost` is rescued both here and in `internal_store_offset`, and I'm not that knowledgable of librdkafka internals.

mensfeld: Yes. Exactly.

PR #1514, opened 2023-07-13
karafka-rdkafka bump

mensfeld:
+ small stability for one spec. It seems new release polls data in pausing in smaller quantities


mensfeld:
Does not fix https://github.com/confluentinc/librdkafka/issues/4312 but nonetheless decent release so worth updating. No new problems introduced.

PR #1567, opened 2023-08-21
Update signing cert

mensfeld:
Old yearly cert expired two days ago. Rotating. @nijikon pinging you via a PM to confirm.


nijikon (approved):
I confirm this is legit.

PR #1576, opened 2023-08-23
Support changing service_name in datadog integration

paneq:
By default datadog will use the service name such as `myservice` which will be the same for web and karafka processes. 

However, it is beneficial to separate those and have  `myservice-karafka` as a separate service name. This allows setting different default operation in Datadog UI and makes analyzing traces much easier.

In https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/ruby/ you can find examples how every integration supports setting service name.

Tested manually in Datadog as well.
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.1.11 (2023-08-23)
 - [Enhancement] Expand the error handling for offset related queries with timeout error retries.
-- [Enhancement] Allow for connection proxy timeouts configuration
+- [Enhancement] Allow for connection proxy timeouts configuration.
 
 ## 2.1.10 (2023-08-21)
 - [Enhancement] Introduce `connection.client.rebalance_callback` event for instrumentation of rebalances.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Karafka framework changelog
 
-## 2.1.11 (Unreleased)
+## 2.1.11 (2023-08-23)
 - [Enhancement] Expand the error handling for offset related queries with timeout error retries.
 - [Enhancement] Allow for connection proxy timeouts configuration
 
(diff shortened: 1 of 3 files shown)


mensfeld:
Not a heavy DD user anymore: would it be beneficial to do the same to the second DD listener?

lib/karafka/instrumentation/vendors/datadog/logger_listener.rb
```diff
@@ -12,11 +12,13 @@ class LoggerListener
           include ::Karafka::Core::Configurable
           extend Forwardable
 
-          def_delegators :config, :client
+          def_delegators :config, :client, :service_name
 
           # `Datadog::Tracing` client that we should use to trace stuff
           setting :client
 
+          setting :service_name
```
mensfeld, line 20: is there a default for this value or will the default be used when nil?

mensfeld (commented):
One question: is this change backwards compatible? could there be an explicit default that would match the current behaviour?

lib/karafka/instrumentation/vendors/datadog/logger_listener.rb
paneq: default when `nil`:  https://github.com/DataDog/dd-trace-rb/blob/afc4a025c7c8717050901f1419b7b21856aa9a2a/lib/datadog/tracing/tracer.rb#L128

>  is this change backwards compatible? could there be an explicit default that would match the current behaviour?

so it's backwards compatible. The only difference is that now `nil` will be passed explicitly when not set.

paneq:
> would it be beneficial to do the same to the second DD listener?

No idea. Tried to look into ddtrace gem codebase to find out, but I couldn't quickly figure it out. My gut feeling is that yes, it should be possible an beneficial.

mensfeld (approved):

mensfeld (approved):

mensfeld:
Retargeting so I can polish it. Once merged to master will be part of the upcoming release since it does not have breaking changes :bowing_man:

Issue #1579, opened 2023-08-24
DLQ + VP without intermediate marking may mark earlier message


mensfeld:
When using DLQ with VP without `mark_as_consumed` on each VP (at least once at the end) may automatically mark earlier message than the last.

PR #1580, opened 2023-08-24
Fix DLQ + VP without intermediate marking may mark earlier message

mensfeld:
close https://github.com/karafka/karafka/issues/1579


mensfeld:
I added some extra specs also to not affected scenarios just for the future.

Issue #1625, opened 2023-09-20
Untangle the routing reference to subscription groups


mensfeld:
At the moment depending on where you are, there are small discrepancies in `#subscription_group` reference. In routing topic this is a string but when operating on the routing tree it is an object. Instrumentation also uses it as an object. So this should be aligned. Those are internal structures so it should not be a problem to change it because no APIs rely on it as a value.

PR #1626, opened 2023-09-20
Untangle routing subscription group reference

mensfeld:
This PR:

- normalizes how we operate with subscription groups from a routing topics perspective to align it with a consumer group perspective. It does NOT change any official APIs of the routing.
- renames the `subscription_group_id` to `subscription_group_name` because we operate on the group name and not final target IDs. Why? Because it is about grouping topics, BUT the final subscription groups ids need to consider static group memberships.
- Adds the `BaseConsumer#pause` reference of subscription group to align with listener client instrumentation.
- Changes the Kafka container from wurstmeister to confluent and updates the version
- Aligns specs because newer Kafka returns more data on the first poll faster and we've relied on this behavior unintentionally in some of the integration specs.
- Introduces a per OS lock that ensures, that we create only one topic at a time without concurrent creations to prevent overload of Kafka
- Introduce code to wait for Kafka which starts faster now and may not be fully ready before specs start to run.

close https://github.com/karafka/karafka/issues/1625


lib/karafka/routing/subscription_groups_builder.rb
```diff
@@ -39,6 +39,13 @@ def call(topics)
           .map { |value| value.map(&:last) }
           .map { |topics_array| Routing::Topics.new(topics_array) }
           .map { |grouped_topics| SubscriptionGroup.new(@position += 1, grouped_topics) }
+          .tap do |subscription_groups|
+            subscription_groups.each do |subscription_group|
+              subscription_group.topics.each do |topic|
+                topic.subscription_group = subscription_group
```
nijikon, line 45: Shouldn't this be `topic.subscription_group_id = subscription_group_id`?

mensfeld: No. Here we assign the final materialized subscription group. Id is an intermediate "notion" of groups when routing are being built.

nijikon (approved):

PR #1635, opened 2023-09-22
Add an integration spec illustrating how virtual partitioner that distributes all to one partition cancels VPs

mensfeld:
close https://github.com/karafka/karafka/issues/1603

on top improves some specs


nijikon (approved):

mensfeld:
@nijikon I had to lower the concurrency  back because the machine was overloaded (GH runners are 2 core single CPU) and some specs were reaching check times (some are time sensitive) because machine was overloaded.

nijikon (approved):

Issue #1679, opened 2023-10-19
Remove `thor` and use homebrew CLI layer


mensfeld:
Our case is not that complex. We have few options for few commands and commands are already abstracted.

Since thor `1.3.0` broke karafka in an unexpected way, it may be worth considering its removal.

PR #1680, opened 2023-10-19
Remove `thor`

mensfeld:
This PR removes thor and switches to just using opt parser provided by ruby.

close https://github.com/karafka/karafka/issues/1679


mensfeld:
@nijikon specs will fail because they rely on karafka-web that still uses thor API. It will be in a separate PR.

nijikon (approved):
LGTM

PR #1728, opened 2023-11-07
Provide time-reliable events polling

mensfeld:
This PR replaces a single queue based consumer client with one that handles consumer and other events independently + provides a tick engine for periodic checking also when locked during processing or long pollings.

It does NOT increase number of threads needed and performance impact should be neglectable.

This PR allows us to ensure we are publishing statistics more or less every 5 seconds and that they are updated despite processing state or any features usage.


nijikon (approved):
LGTM 🚀

Issue #871, opened 2022-06-10
Support dedicated worker batch for subscription groups


mensfeld:
We could add an expansion to our routing DSL to build named worker polls and just assign them to consumer groups / subscription groups (not to topics):

```ruby
# Auto created if none based on the concurrency
workers_batch(:general, concurrency: 5)
workers_batch(:lrj, concurrency: 3)

# and then when defining CGs, SGs:

consumer_group :xyz do
  workers_batch :general
  
  topic :x do
    consumer LolConsumer
  end
end
```

mensfeld:
Few notes:

- We should allow for fixed value as well as ranges
- Ranges should be used when assignment is big (multi-partition)
- We should scale down given group if it is not assigned not to keep worker threads that will never do anything
- Max concurrency should define the top limit
- By default we should always keep 1 thread allocation for each worker group so we never end up starving a group when upscale would be needed.

mensfeld:
Now that I think of it and with some consultation, I think this should work differently. We should have a scheduler API that would allow us to manage the worker poll that we have in terms of work being done. That way, we could have a single workers poll (the way it is now) but that would be able to process different jobs independently. Like a multi-pre-queue model.

mensfeld:
It has been merged.

PR #1743, opened 2023-11-14
Stateful state-aware scheduling API

mensfeld:
This is a LOW LEVEL API for building state-aware schedulers.

This PR introduces a low-level API for building advanced multi-level schedulers.

Initially, we've wanted to introduce a separate workers' polls management for work distribution; however, due to the nature of work and the fact that it is often dynamically assigned, such an approach would drastically limit us. For example user would be forced to assign idle threads for LRJ (assuming wanted to have 2 workers) to all processes even if they would not have any LRJs assignments.

This API mitigates this by allowing to introduce a pre-buffer with active work distribution on ticks. With that we can make smart state based decisions on final work scheduling depending on the resources availability and assignmnets.

No performance regression.
There are no breaking changes since we only alter the internal scoped configuration that the end users should not change.

part of https://github.com/karafka/karafka/issues/871

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Karafka framework changelog
 
 ## 2.2.13 (Unreleased)
+- **[Feature]** Introduce low-level extended Scheduling API for granular control of schedulers and jobs execution [Pro].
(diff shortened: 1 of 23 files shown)


nijikon (commented):
LGTM 🚀

Issue #1752, opened 2023-11-18
Split before_schedule in the executor to before_consume_schedule etc


mensfeld:
We cannot use the same schedule hook for executor that executes different jobs stages as they have different flow and expectations.

PR #1753, opened 2023-11-18
Split single before schedule to per-action one

mensfeld:
We used to have a generic `before_schedule` on the executor, but this is insufficient for job tracking with advanced schedulers. On top of this, it was not clear what would be the proper post-schedule flow. This PR replaces it with dedicated methods.

On top of that, I removed error catching for critical non-user-related places where errors should be considered critical and not silenced.

close https://github.com/karafka/karafka/issues/1752

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,9 @@
 # Karafka framework changelog
 
 ## 2.2.14 (Unreleased)
+- [Change] Replace single `#before_schedule` with appropriate methods and events for scheduling various types of work. This is needed as we first of all may run different framework logic on those and second for accurate jobs tracking with advanced schedulers.
 - [Change] Rename `before_enqueue` to `before_schedule` to reflect what it does and when (internal).
+- [Change] Remove not needed error catchers for strategies code. This code if errors, should be considered critical and should not be silenced.
 
 ## 2.2.13 (2023-11-17)
 - **[Feature]** Introduce low-level extended Scheduling API for granular control of schedulers and jobs execution [Pro].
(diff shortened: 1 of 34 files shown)


lib/karafka/processing/jobs/base.rb
```diff
@@ -22,9 +22,11 @@ def initialize
           @non_blocking = false
         end
 
-        # When redefined can run any code prior to the job being enqueued
+        # When redefined can run any code prior to the job being scheduled
         # @note This will run in the listener thread and not in the worker
-        def before_schedule; end
+        def before_schedule
+          raise NotImplementedError, 'Please implement in a subclass'
```
nijikon, line 28: `Implement in a subclass`

CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 # Karafka framework changelog
 
 ## 2.2.14 (Unreleased)
+- [Change] Replace single `#before_schedule` with appropriate methods and events for scheduling various types of work. This is needed as we first of all may run different framework logic on those and second for accurate jobs tracking with advanced schedulers.
```
nijikon, line 4: Replace single `#before_schedule` with appropriate methods and events for scheduling various types of work. This is needed as we may run different framework logic on those and, second, for accurate job tracking with advanced schedulers.

lib/karafka/processing/jobs/base.rb
mensfeld: Would have to change in all the places where this message is. Outside of the scope but feel free to create an issue :)

nijikon: Fair enough.

nijikon (approved):

Issue #1751, opened 2023-11-18
Assignments tracker


mensfeld:
As part of instrumentation, it would be good to have ability to get current assignments.

PR #1760, opened 2023-11-21
Assignments Tracker

mensfeld:
Tracker that allows us to get current assignments. Can be useful when using advanced scheduling APIs

close https://github.com/karafka/karafka/issues/1751


lib/karafka/instrumentation/assignments_tracker.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+module Karafka
+  module Instrumentation
+    # Keeps track of active assignments and materializes them by returning the routing topics
```
nijikon, line 5: `moment'` -> `moment`

nijikon (approved):

PR #1775, opened 2023-12-05
Better scheduler api

mensfeld:
While writing docs for this feature, I noticed that it is fundamentally not thread-safe. We should not expose if possible direct APIs for end users that are complex. This PR updates that and allows for both thread-safe as well as fast non-locked schedulers that are stateless.


nijikon (commented):
Nice one 💜

PR #1783, opened 2023-12-10
Periodic Jobs

mensfeld:
This code introduces periodic jobs that can run on each poll for consumers needing such behavior (with a tick_interval limitation not to overload). It is useful when we do window processing, cron jobs, and advanced flows that are not always messages.


lib/karafka/pro/processing/schedulers/base.rb
```diff
@@ -106,18 +88,23 @@ def on_schedule_idle(jobs_array)
             end
           end
 
-          # Schedules the idle jobs.
+          # Runes the periodic jobs scheduling flow under a mutex
```
nijikon, line 91: Runs

nijikon (approved):
Another epic feature, love it 🚀

PR #1817, opened 2023-12-24
alias lrj as non blocking

mensfeld:
This PR aliases LRJ via `non_blocking` to allow for indication in routing that LRJ is set to avoid blocking and not to run LRJ.

While conceptually it uses the same code, the expression "non blocking" and "long-running" have different indication


nijikon (approved):
LGTM 🎄

PR #1829, opened 2023-12-29
Offsets metadata management

mensfeld:
This PR adds support for offset metadata management in Karafka.


lib/karafka/connection/client.rb
```diff
@@ -353,12 +362,14 @@ def events_poll(timeout = 0)
       # The offset field of each requested partition will either be set to stored offset or to
       # -1001 in case there was no stored offset for that partition.
       #
+      # @param [Rdkafka::Consumer::TopicPartitionList] tpl for which we want to get committed
```
nijikon, line 365: `# @param tpl [Rdkafka::Consumer::TopicPartitionList] for which we want to get committed`

PR #1837, opened 2024-01-06
Improved Periodics

mensfeld:
I noticed that periodics, by default, would run when the partition is paused and on errors. I introduced two new settings to them that allow for configuration of it. The defaults are as follows:

- on recovery (retry flows) periodics will NOT run
- on pause periodics WILL run unless it is LRJ because for LRJ pause is applied during processing. This provides an "out of the box" way of ensuring that LRJ jobs won't run with periodics unless explicitly requested.
-


nijikon (approved):
LGTM

PR #1840, opened 2024-01-07
API for scaling listener

mensfeld:
This PR provides new API for managing listeners. It also removes old patches for librdkafka (no longer needed) and improves shutdown by moving the coordination logic to the manager. It also removes global state presence from listeners so now only manager can manage them (and server on forceful shutdown if needed) and provides several other improvements that rely on recent librdkafka advances.

It does implement the pro manager.


nijikon (approved):
LGTM 🚀

PR #1882, opened 2024-02-05
Move running status management to runner

mensfeld:
This PR moves run! status trigger to runner to be able to start it in forks once we have them, so the status is not switched until forking kicks in.

### DIFF
--- a/spec/lib/karafka/server_spec.rb
+++ b/spec/lib/karafka/server_spec.rb
@@ -105,14 +105,12 @@ RSpec.describe_current do
 
   describe '#start' do
     before do
-      allow(Karafka::App).to receive(:run!)
       allow(runner).to receive(:call)
 
       server_class.start
     end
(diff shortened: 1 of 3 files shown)


spec/lib/karafka/server_spec.rb
```diff
@@ -105,14 +105,12 @@
 
   describe '#start' do
     before do
-      allow(Karafka::App).to receive(:run!)
```
nijikon, line 108: Should we move this to the https://github.com/karafka/karafka/blob/master/spec/lib/karafka/runner_spec.rb?

mensfeld: I do not see any reason of adding this there, it is fully covered by the integration suite and it's a leftover of old times when we didn't have them at all

nijikon (approved):

Issue #1897, opened 2024-02-14
Introduce intermediate errors tracker for complex recovery flows


mensfeld:
At the moment the error information is "lost" when moving to recovery. We should keep track of last 100 errors for recovery flows and only clean on another clean run post recovery including complex recovery flows with during-recovery errors.

Issue #1896, opened 2024-02-14
Store last error that occurred details in the consumer for retry action referencing


mensfeld:
useful for error driven flows

PR #1899, opened 2024-02-14
track execution errors during the recovery flow

mensfeld:
This PR adds a tracker that allows us to get the errors that occurred during the execution until no error via retry or DLQ dispatch.

Additionally this PR moves the `Coordinator` related sub-components used only internally via the coordinator to `Coordinators` namespace.

close https://github.com/karafka/karafka/issues/1897
close https://github.com/karafka/karafka/issues/1896


nijikon (commented):
LGTM 👍

PR #1924, opened 2024-02-22
Add `use_distributions` setting to Datadog::MetricsListener

Aerdayne:
Currently, histogram metrics are emitted as histograms in the Datadog listener.
This PR introduces an option to emit these metrics as distributions instead. This makes the calculated percentiles more accurate in case there are multiple consumers running on different hosts (or rather when multiple DD agents are being used).
More info can be found in the [DD documentation](https://docs.datadoghq.com/metrics/types/?tab=distribution#metric-types).

### DIFF
--- a/lib/karafka/instrumentation/vendors/datadog/metrics_listener.rb
+++ b/lib/karafka/instrumentation/vendors/datadog/metrics_listener.rb
@@ -15,7 +15,7 @@ module Karafka
           extend Forwardable
 
           def_delegators :config, :client, :rd_kafka_metrics, :namespace,
-                         :default_tags, :use_distributions
+                         :default_tags, :distribution_mode
 
           # Value object for storing a single rdkafka metric publishing details
           RdKafkaMetric = Struct.new(:type, :scope, :name, :key_location)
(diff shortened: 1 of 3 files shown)


mensfeld:
:wave: I'll pick it up in the upcoming days :pray:

mensfeld:
@Aerdayne this looks good. I only have a conceptual remark and I am happy to apply it if you do not want or do not have time:

I do not like `use_distribution` config. I would rather have `distribution_mode` and have it set o `:histogram`. Then we could use the `#distribution` method like other metric types and underneath use context awareness to pass to the client. This would in my opinion be a bit more readable because it would not introduce the public send

mensfeld (commented):
comment in the comments ;)

Aerdayne:
So you mean something like this?

```ruby
case distribution_mode
in :histogram
  histogram(...)
  # ...
in :distribution
  distribution(...)
  # ...
end
```

mensfeld:
> So you mean something like this?
> 
> ```ruby
> case distribution_mode
> in :histogram
>   histogram(...)
>   # ...
> in :distribution
>   distribution(...)
>   # ...
> end
> ```

exactly :)

mensfeld:
Thank you for the PR btw. I know karafka DX from the contributor perspective is shit at the moment :(

mensfeld (approved):

mensfeld:
@Aerdayne  I applied a small refactor on top of your code. Would you mind taking a look? https://github.com/karafka/karafka/pull/1928 I cannot assign you as the reviewer so pinging you here

Issue #1809, opened 2023-12-18
Consider expanding the iterator with ability to mark offsets


mensfeld:
it could then be used for rake etc

PR #1949, opened 2024-03-05
introduce marking as consumed in the iterator

mensfeld:
close https://github.com/karafka/karafka/issues/1809

this PR also moves some of the generic abstraction to rdkafka proxy and caches the proxy wrapper not to create it when not needed.

### DIFF
--- a/lib/karafka/pro/iterator.rb
+++ b/lib/karafka/pro/iterator.rb
@@ -151,7 +151,6 @@ module Karafka
       def mark_as_consumed!(message)
         mark_as_consumed(message)
         @current_consumer.commit_offsets(async: false)
-        @stored_offsets = true
       end
 
       private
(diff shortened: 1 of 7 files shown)


lib/karafka/pro/iterator.rb
```diff
@@ -127,6 +130,30 @@ def stop_partition(name, partition)
...
+        @stored_offsets = true
+      end
+
+      # Marks given message as consumed and commits offsets
+      #
+      # @param message [Karafka::Messages::Message] message that we want to mark as processed
+      def mark_as_consumed!(message)
+        mark_as_consumed(message)
+        @current_consumer.commit_offsets(async: false)
+        @stored_offsets = true
```
nijikon, line 154: This is already set via `mark_as_consumed` method

nijikon (approved):

Issue #1951, opened 2024-03-05
Remove concept of consumer mappers for 2.4


mensfeld:
Similar to previous topic mappers this creates a lot of noise and is extremely complex to handle.

Users should build their own naming conventions and not rely on artificially expanded once that create noise.

This is a breaking change but a needed one.

PR #1952, opened 2024-03-05
drop CG mappers and introduce read_lags

mensfeld:
note: I do not update the version to 2.4.0 yet to make all specs work with ecosystem gems. Will bump when releasing RC1.

close https://github.com/karafka/karafka/issues/1951
close https://github.com/karafka/karafka/issues/1950


nijikon:
LGTM

PR #1969, opened 2024-03-11
ensure k8s and swarm liveness works with multiplexing

mensfeld:
close https://github.com/karafka/karafka/issues/1956
close https://github.com/karafka/karafka/issues/1970


lib/karafka/pro/swarm/liveness_listener.rb
```diff
@@ -99,10 +99,32 @@ def on_statistics_emitted(_event)
 
             current_status = status
 
+            p current_status
```
nijikon, line 102: Remove debug

PR #1978, opened 2024-03-17
Introduce Admins low-level configs API

mensfeld:
This does not yet cover declaratives. Just introduces basic low level admin API for configs management similar to ACLs.


mensfeld:
@nijikon PRs for web will fail because web needs deps update that will be possible only when alpha1 is released.

nijikon (commented):
LGTM 🚀

PR #1990, opened 2024-03-22
make #enqueue_all output compatible with ActiveJob.perform_all_later

oozzal:
For any ActiveJob adapter, `#enqueue_all` method should return the total number of jobs enqueued.

So that `enqueued_count` set by `instrument_enqueue_all` is as expected: 
https://github.com/rails/rails/blob/main/activejob/lib/active_job/enqueuing.rb#L17
https://github.com/rails/rails/blob/main/activejob/lib/active_job/instrumentation.rb#L10

Otherwise, it fails here: https://github.com/rails/rails/blob/main/activejob/lib/active_job/log_subscriber.rb#L68


mensfeld (approved):

mensfeld:
Amazing. Thank you. It will go with the upcoming release :pra

Don't worry about specs failing, still didn't figure out how to run it fully on forks.

Issue #1994, opened 2024-03-25
Allow using `mark_as_consumed!` in the DLQ dispatch instead of default


mensfeld:
currently we use `mark_as_consumed` but we could have it differentiated based on the user settings. While using async is recommended, some users may actually want to ensure that DLQ dispatch does not end up being retried upon process kill.

PR #1995, opened 2024-03-25
improve dlq dispatch warranties

mensfeld:
this PR allows for selection whether we want to mark as consumed async or sync post dispatch on dlq and whether we want to dispatch sync or async (in transactions it is irrelevant)

close https://github.com/karafka/karafka/issues/1994


nijikon (commented):
LGTM

PR #1999, opened 2024-03-26
replace subscribe best offset position with admin api

mensfeld:
close https://github.com/karafka/karafka/issues/1378

this PR shaves over 11 minutes from running CI. Since we run in parallel (3 processes as far as I remember) the total gain should be around 3-4 minutes.


nijikon (approved):
LGTM

Issue #2006, opened 2024-03-27
Key Deserialization


dorner:
## Is your feature request related to a problem? Please describe.

I'm looking at using Karafka as the new backend for [Deimos](https://github.com/flipp-oss/deimos). One important feature that Deimos has is the ability to Avro-encode/decode keys as well as messages. This doesn't seem to be available on Karafka right now.

Serialized keys are important because we work with a number of Scala services, which understand the data better if the keys are encoded. A number of messages have multi-field keys, for example.

## Describe the solution you'd like

The ability to specify both a `deserializer` and a `key_deserializer` on a consumer. Essentially it should follow the same logic as the `deserializer` but act on keys instead of messages.

## Describe alternatives you've considered

We'd have to manually deserialize the keys ourselves in a base consumer. This is doable but eliminates a lot of the fluidity of relying on the framework to do it for us.

## Additional context

All our topics currently have Avro-encoded keys - this would be a soft blocker to moving to Karafka (as I said, there is a workaround).

Definitely willing to give a PR a shot if there are other priorities right now.

mensfeld:
So you basically deserialize each header independently right?

dorner:
This is for the message key, not the headers.

mensfeld:
aaa shit sorry, I misread it :facepalm:

mensfeld:
although I think we could expand it and update the routing to do this:
```ruby
deserializers(
  payload: ...,
  key: ...,
  headers: ...
)
```

and alias current deserializer as the payload one

mensfeld:
Yeah this is a good idea. I should be able to ship it in the upcoming release as it does not require a major work.

dorner:
Amazing! Thanks so much for the prompt reply. Please let me know if I can help.

mensfeld:
Nah all good. I already have a poc :rofl: - just need to finish the oauth token work (in the upcoming days) and I should be able to ping you with the PR next week.

mensfeld:
Yeah all of my integration specs work. It's lazy deserialized in the same way as payload with default to doing nothing with both key as well as heders.

PR #2007, opened 2024-03-27
support key and header deserializers

mensfeld:
close https://github.com/karafka/karafka/issues/2006


mensfeld:
some specs are expected to fail because testing and web are not yet aligned

PR #2011, opened 2024-03-29
stabilize spec and update changelog

mensfeld:


spec/integrations/pro/consumption/direct_assignments/transactions_support_spec.rb
```diff
@@ -23,6 +23,8 @@ def consume
 draw_routes do
   topic DT.topic do
     consumer Consumer
+    # Not a VP case but we ant to stabilize number of messages to prevent randomness
```
nijikon, line 26: `want`

PR #2056, opened 2024-04-23
Give up on running .0

mensfeld:
Similar to rdkafka, this should not be our policy when this can be upgraded easily.


nijikon (approved):
LGTM

PR #2057, opened 2024-04-23
Protect rdkafka thread callbacks flow from errors

mensfeld:
same case as in waterdrop

here we do have a recovery on the listener layer, which would recover the flow, but to be consistent with waterdrop and how we handle callback execution, this was added close to callback handlers as well.

ref https://github.com/karafka/karafka-rdkafka/issues/90


nijikon (approved):
LGTM

Issue #2064, opened 2024-04-26
Support migrating via aliases and plan with aliases usage


mensfeld:
At the moment when we run plan or migrate, aliases attributes are not being used.

mensfeld:
For example such fully migrated setup:

```ruby
      config(
        partitions: 5,
        'retention.ms': 10_000,
        'segment.ms ': 10_000,
        'delete.retention.ms': 10_000,
        'max.compaction.lag.ms': 10_000,
        'min.cleanable.dirty.ratio': 0.0,
        'log.cleaner.max.compaction.lag.ms': 100,
        'cleanup.policy': 'compact,delete'
      )
```

is anyhow scheduled for planning:

```bash
Following topics will have configuration changes:

  ~ default:
    + segment.ms : "" => "10000"
    + log.cleaner.max.compaction.lag.ms: "" => "100"

  ~ visits:
    + segment.ms : "" => "10000"
    + log.cleaner.max.compaction.lag.ms: "" => "100"
```

mensfeld:
interestingly enough migrate on this plan gives:
```bash
Skipping because there are no configurations to align.
```

PR #2065, opened 2024-04-26
Support migrating via aliases and plan with aliases usage

mensfeld:
close https://github.com/karafka/karafka/issues/2064

I also noticed that dirs have same name as specs prefixes, thus I renamed them.


nijikon (approved):
LGTM

PR #2078, opened 2024-05-06
Extra spec illustrating that pausing won't bypass max.poll

mensfeld:
This spec ensures that post-pause locking still works.


nijikon (approved):
LGTM

PR #2084, opened 2024-05-07
more seek specs

mensfeld:
This PR adds more specs to seeking + removes the redundant `seek_` in the name since directory is already `seek/`


nijikon (approved):
LGTM

Issue #2090, opened 2024-05-09
Add a pristine spec for only waterdrop to ensure its internal system gems are all as expected


mensfeld:
ref https://github.com/karafka/waterdrop/pull/485

PR #2091, opened 2024-05-10
Ensure pristine WaterDrop can run

mensfeld:
close https://github.com/karafka/karafka/issues/2090

discovered here: https://github.com/karafka/waterdrop/pull/485

I tested it against `2.7.1` and it crashed. So works as expected.


nijikon (commented):
LGTM

PR #2095, opened 2024-05-10
Support variants and alternative producers in ActiveJob

mensfeld:
close https://github.com/karafka/karafka/issues/2094


nijikon (approved):
LGTM

PR #2108, opened 2024-05-14
Config rework

dorner:
This isn't finished yet. The idea here is to use `:not_given` as all config parameters, construct the config proactively, and only consider it a "read" operation if no parameters are given at all.

periodic\_job is a bit more in-depth, so I'm leaving off on it for now - I just wanted to validate this approach first and get feedback. Also not sure if there's additional configuration I'm not aware of, I was looking at the topic stuff.

The main advantage here is the ability to make changes to individual parts of each config rather than blowing away the whole thing at once. It also opens up the ability to redefine / reconfigure things dynamically if needed - right now, you can only configure any feature once, either by explicit setting or by default, and never again.

@mensfeld


dorner:
oh I also added a `redraw` method for routes - I can remove that from this PR if it's doing too much.

dorner:
I couldn't run all specs locally. There are a couple that failed where I changed the spec. In particular, it was testing that if it set `active` first to true, and then to false, that the result would `true`. This seems wrong to me. The new config correctly (IMO) returns `false` in this case.

Issue #2085, opened 2024-05-07
[v2.4.0] Unable to consume messages (only on ConfluentCloud)


hbontempo-cw:
First of all: Thanks for the great work. I love to use karafka 😃 

## Expected behavior

After upgrading Karafka from v2.3.4 to v2.4.0 everything was expected to keep working since none of the features remove are being used for the application.

## Actual behavior

After upgrading Karafka from v2.3.4 to v2.4.0 the karafka server stopped consuming messages.
Messages are still been sent normally.

More strange:
This problem **only** happens on Kafka hosted in ConfluentCloud.
Using a local kafka (docker) everything works fine.

## Steps to reproduce the problem

```ruby
# karafka.rb

class KarafkaApp < Karafka::App
  setup do |config|
    config.client_id = 'test'
    config.kafka = {
      :"bootstrap.servers"=>"[REDACTED]",
      :"security.protocol"=>"SASL_SSL",
      :"sasl.mechanism"=>"PLAIN",
      :"sasl.username"=>"[REDACTED]",
      :"sasl.password"=>"[REDACTED]",
      :"compression.codec"=>"gzip",
      :"compression.level"=>12}
    config.logger = Rails.logger
    config.concurrency = 5
  end

  routes.draw do
    consumer_group :test_subscriber do
      topic('test-topic') { consumer TestConsumer }
    end
  end
end

```

Producing messages works fine:
```sh
$ RAILS_ENV=staging bundle exec rails c
D, [2024-05-07T15:04:43.078861 #34370] DEBUG -- : [dotenv] Set [REDACTED]
I, [2024-05-07T15:04:43.086859 #34370]  INFO -- : [dotenv] Loaded .env.staging
D, [2024-05-07T15:04:43.370596 #34370] DEBUG -- sentry: Sentry HTTP Transport will connect to [REDACTED]
Loading staging environment (Rails 7.1.3.2)
irb(main):001> Karafka.producer.produce_sync(payload: '{"a": 1}', topic: 'test-topic')
=> 
#<Rdkafka::Producer::DeliveryReport:0x000000012dd3d758
 @error=nil,
 @label=nil,
 @offset=2766,
 @partition=1,
 @topic_name="test-topic">

```

But consuming messages is not working:
```sh
$ RAILS_ENV=staging bundle exec karafka s
I, [2024-05-07T15:04:43.078606 #34370]  INFO -- : [dotenv] Loaded .env.staging.local
[...]

PR #2109, opened 2024-05-16
Add some more error codes to early report

kidlab:
Recently, after upgrading to Karafka 2.4.0 by mistake, I encountered this issue https://github.com/karafka/karafka/issues/2085. There was no error reported on staging and production servers, the Karafka process appeared to be up and running but didn't consume any messages. We had to enable the debug mode in local host to understand what went wrong.

So for errors like `topic_authorization_failed`, `group_authorization_failed` and `cluster_authorization_failed`, `rdkafka` will raise the error only on the first poll, from the second poll tick, the message will be always null and make it look good, but in fact, cause the entire `poll` return null result without any error: https://github.com/karafka/rdkafka-ruby/blob/2900f6aa0d281ef8c12918c076b1cbab3f0d86fe/lib/rdkafka/consumer.rb#L530-L531

Therefore, in this PR, I'm adding those codes to the list of early-report errors.

### DIFF
--- a/lib/karafka/connection/client.rb
+++ b/lib/karafka/connection/client.rb
@@ -553,6 +553,12 @@ module Karafka
           early_report = true
         when :transport # -195
           early_report = true
+        when :topic_authorization_failed # 29
+          early_report = true
+        when :group_authorization_failed # 30
+          early_report = true
+        when :cluster_authorization_failed # 31
+          early_report = true
         # @see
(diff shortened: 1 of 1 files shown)


mensfeld:
@kidlab I will take a look once CLA is signed :pray:

kidlab:
@mensfeld Done :heavy_check_mark:

mensfeld:
merged. I will initiate 2.4.3 in GH, and this should go live in 1-2 weeks with a few other small enhancements I have in plans. Thank you and sorry for the trouble.

Issue #2127, opened 2024-06-05
Introduce VPs partitioner reducer


mensfeld:
Currently VPs always use `%` on `to_s.sum` to distribute work. While this is most of the cases good, there are cases when this may not allocate messages to all partitions. Especially on independent messages. We should have a new entity that is responsible for taking partitioner result and building the distribution.

PR #2129, opened 2024-06-05
Provide ability to use custom VP reducers

mensfeld:
This PR extracts the default reducer code to a customizable reducer, allowing for the usage of custom reducers for better (wider) mapping of the virtual partition keys.

As shown in the included specs, the default reducer is good but may not cover all people's cases due to the nature of their keys or concurrency settings. This is compensated by giving them the ability to write their own.

close https://github.com/karafka/karafka/issues/2127


nijikon (approved):
LGTM

PR #2139, opened 2024-06-16
Propagate fatal errors faster

mensfeld:
close https://github.com/karafka/karafka/issues/1921

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Karafka framework changelog
 
+## 2.4.4 (Unreleased)
+- [Fix] Static membership fencing error is not propagated explicitely enough.
+
 ## 2.4.3 (2024-06-12)
 - [Enhancement] Allow for customization of Virtual Partitions reducer for enhanced parallelization.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Karafka framework changelog
 
+## 2.4.4 (Unreleased)
+- [Fix] Static membership fencing error is not propagated explicitely enough.
```
nijikon, line 4: `Static membership fencing error is not propagated explicitly enough`

lib/karafka/connection/client.rb
```diff
@@ -559,6 +559,9 @@ def poll(timeout)
           early_report = true
         when :cluster_authorization_failed # 31
           early_report = true
+        # This can happen for many reasons, including issues with static membership being fenced
+        when :fatal
+          early_report = true # -150
```
nijikon, line 564: What is `-150` here? Shouldn't it be on the `fatal` line as above?

nijikon (approved):

PR #2143, opened 2024-06-17
update specs to match core 2.4.1 and require it

mensfeld:
This PR ensures we operate correctly with updated core 2.4.1. The RC requirement will be lifted after non RC is released.


nijikon (approved):
LGTM

PR #2145, opened 2024-06-17
allow for usage of karafka-rdkafka 0.16.0

mensfeld:
The `>= 0.16.0` is just to test. We will keep it open to 0.15 for a while to allow users to stay on `librdkafka` `2.3` if needed.


mensfeld:
@nijikon I will update this PR with non-RC karafka-rdkafka when al checks are :green_heart:

Issue #2158, opened 2024-07-01
Print more extensive debug info on forceful shutdown


mensfeld:
Since we know what is still active (workers, listeners) we should be able to print this during forceful shutdown to improve debugibility.

PR #2159, opened 2024-07-01
improve logger listener instrumentation and shutdown

mensfeld:
similar to waterdrop we lower the noise of time reporting
additionally I've also improved the forceful shutdown to prevent a situation when we hang on listener shutdown despite forceful.

close https://github.com/karafka/karafka/issues/2158


nijikon (commented):
LGTM

PR #2222, opened 2024-08-14
Bump ActiveSupport

mensfeld:


.github/workflows/ci.yml
```diff
@@ -77,7 +77,6 @@ jobs:
           - '3.3'
           - '3.2'
           - '3.1'
-          - '3.0'
```
nijikon, line 80: @mensfeld Does this mean that we will drop support for `3.0`?

mensfeld: no, it will still run full e2e integrations just not the rspec in oss because activesupport ain't operating with it and it's not worth heavy engineering

nijikon (approved):
LGTM

Issue #2167, opened 2024-07-08
Heavily loaded, multi-threaded consumer process hangs on Rails code reload in dev


mensfeld:
Only in dev, heavily utilized, multiplexed consumer process can hang on rails locks when code reloading.

PR #2232, opened 2024-08-21
Draft/WIP/Idea: Wrap worker/job execution with rails reloader or executor

fusion2004:
This is a stab at removing the manual code reloading and instead wrapping execution (if integrated with a rails app) with the rails executor or reloader (depending on whether consumer_persistence is enabled or not). I was having some wild reloading issues in development that I could not seem to fix... but this does it!

This is how sidekiq handles this as well: https://github.com/sidekiq/sidekiq/blob/ed8b217d32f026acc1fae7221dca18d1baa86d11/lib/sidekiq/processor.rb#L135-L147

There seems to be other important things handled by the executor we likely want! (Query cache enabling/disabling, connection management -- I didn't try it out, but possibly we won't need the other initializer that clears AR connections on worker completion?) https://guides.rubyonrails.org/threading_and_code_execution.html#wrapping-application-code

### DIFF
--- a/lib/karafka/processing/worker.rb
+++ b/lib/karafka/processing/worker.rb
@@ -17,6 +17,9 @@ module Karafka
     #                code. This can be used to unlock certain resources or do other things that are
     #                not user code but need to run after user code base is executed.
     class Worker
+      include Helpers::ConfigImporter.new(
+        external_worker_executor: %i[internal processing external_worker_executor]
+      )
       include Helpers::Async
 
       # @return [String] id of this worker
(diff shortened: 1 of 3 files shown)


lib/karafka/processing/worker.rb
```diff
@@ -52,19 +55,21 @@ def process
           Karafka.monitor.instrument('worker.process', instrument_details)
 
           Karafka.monitor.instrument('worker.processed', instrument_details) do
-            job.before_call
+            external_worker_executor.call do
```
fusion2004, line 58: I'm not sure if this is the best place to integrate it, but it works! Originally, I had it wrapping this entire method including the `job = @jobs_queue.pop` line, and that caused what I assume are threadsafety errors, something about adding to a hash we are iterating.

lib/karafka/railtie.rb
```diff
@@ -136,6 +114,18 @@ class Railtie < Rails::Railtie
           end
         end
       end
+
+      initializer 'karafka.configure_worker_external_executor' do |app|
+        app.config.after_initialize do
```
fusion2004, line 119: I don't love putting this in an `after_initialize` callback, but I needed the `consumer_persistence` value after loading the app's boot file.

mensfeld:
Hey. Great timing! This is actually something I've been exploring due to https://github.com/karafka/karafka/issues/2167. I will try to wrap my head around this in the upcoming days.

lib/karafka/processing/worker.rb
mensfeld: that may be expected. The queue is multi-threaded and locks from multiple threads so I can expect it may cause some issues to the reloaded

lib/karafka/setup/config.rb
```diff
@@ -285,6 +285,8 @@ class Config
           setting :expansions_selector, default: Processing::ExpansionsSelector.new
           # option [Class] executor class
           setting :executor_class, default: Processing::Executor
+          # option external_worker_executor [proc] callable object that will be used to wrap execution of a job
+          setting :external_worker_executor, default: proc { |&block| block.call }
```
mensfeld, line 289: not sure if I would call it `executor` per se - maybe more of a wrapper and I would set it to `nil` for defaults and not use it if not needed. That way we don't have to have a second nesting in this critical (but I can do it myself)

mensfeld (approved):

mensfeld:
@fusion2004 I retargeted your PR, cleaned it, added few remarks and here's the ready work: https://github.com/karafka/karafka/pull/2233

If it works for you (it does for me) I think it should be good to merge :)

PR #2233, opened 2024-08-22
Wrap worker/job execution with rails reloader or executor

mensfeld:
continuation of https://github.com/karafka/karafka/pull/2232

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.4.9 (Unreleased)
 - **[Feature]** Provide Kafka based Recurring (Cron) Tasks.
+- [Enhancement] Wrap worker work with Rails Reloader/Executor (fusion2004)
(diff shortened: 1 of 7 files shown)


mensfeld:
@fusion2004 technically user code should be called only in the `job.call`, thus I simplified it.

Checking the boolean is 10-15x faster than running a proc dummy chain.

mensfeld:
I also renamed it to reflect what it does. It does not wrap ALL worker work thus the name seemed not to be perfect. I align it and other pieces of job flow are not part of user code, thus this should be enough. Works on my machine well. Lets see the test suite and waiting on your checks.

fusion2004 (approved):
Yep, working great for me as well!

I wasn't a huge fan of the proc dummy chain either, but I was borrowing that approach from sidekiq. 👍

mensfeld:
Merged. I should wrap this release in a few days.

PR #2241, opened 2024-08-23
Config rework (#2108)

mensfeld:
continuation of the amazing work from here: https://github.com/karafka/karafka/pull/2108

### DIFF
--- a/lib/karafka/pro/routing/features/dead_letter_queue/topic.rb
+++ b/lib/karafka/pro/routing/features/dead_letter_queue/topic.rb
@@ -21,10 +21,11 @@ module Karafka
             # @param strategy [#call, nil] Strategy we want to use or nil if a default strategy
             # (same as in OSS) should be applied
             # @param args [Hash] OSS DLQ arguments
-            def dead_letter_queue(strategy: nil, **args)
-              return @dead_letter_queue if @dead_letter_queue
+            def dead_letter_queue(strategy: Karafka::Routing::Default.new, **args)
+              orig = super(**args)
+              return orig if Karafka::Routing::BaseConfig.all_defaults?(strategy)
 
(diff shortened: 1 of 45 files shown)


mensfeld:
I'm going to put this here for reference:

```
class Matcher
  def initialize
    @applications = []
  end

  def replay_on(topic_node)
    @applications.each do |method, kwargs|
      if kwargs.is_a?(Hash)
        ref = topic_node.public_send(method)

        kwargs.each do |arg, val|
          if ref.respond_to?("#{arg}=")
            ref.public_send("#{arg}=", val)
          else
            if method == :config
              ref.details.merge!(kwargs)
            else
              raise 'No idea if such case exists, if so, similar handling as config'
            end
          end
        end
      end

      if kwargs.is_a?(Array) && kwargs.size == 1
        if topic_node.respond_to?("#{method}=")
          topic_node.public_send(:"#{method}=", kwargs.first)
        else
          topic_node.public_send(method, *kwargs)
        end
      end
    end
  end

  def method_missing(m, *args, **kwargs)
    if args.empty?
      @applications << [m, kwargs]
    else
      @applications << [m, args]
    end
  end
end

DEFAULTS = Matcher.new

module Builder
  def defaults(&block)
    DEFAULTS.instance_eval(&block) if block
  end
end

module ConsumerGroup
  def topic=(name, &block)
    k = Matcher.new
    t = super(name)
    k.instance_eval(&block) if block
    DEFAULTS.replay_on(t)
    k.replay_on(t)
  end
end

Karafka::Routing::Builder.prepend Builder
Karafka::Routing::ConsumerGroup.prepend ConsumerGroup

```

mensfeld:
Closing due to the solution mentioned above that solves this issue without a whole routing reorganization.

PR #2242, opened 2024-08-23
Fix deprecated appsignal warnings

hieuk09:
Fix https://github.com/karafka/karafka/issues/2238


mensfeld:
ref https://github.com/appsignal/appsignal-ruby/blob/main/CHANGELOG.md#deprecated-3

PR #2257, opened 2024-09-02
Fix deprecation warning in `lib/karafka/railtie.rb`

heka1024:
I found this deprecation warning when ready to Rails 7.2 in my application.

> ActiveSupport::DeprecationException: DEPRECATION WARNING: `clear_active_connections!` currently only applies to connection pools in the current role (`writing`). In Rails 7.2, this method will apply to all known pools, regardless of role. To affect only those connections belonging to a specific role, pass the role name as an argument. To switch to the new behavior, pass `:all` as the role name. (called from loop at <internal:kernel>:187) (ActiveSupport::DeprecationException)

We can suppress this warning by passing `:all` to argument.

### DIFF
--- a/lib/karafka/railtie.rb
+++ b/lib/karafka/railtie.rb
@@ -77,7 +77,7 @@ if Karafka.rails?
             # The change technically happens in 7.1 but 7.0 already supports this so we can make
             # a proper change for 7.0+
             if rails7plus
-              ActiveRecord::Base.connection_handler.clear_active_connections!
+              ActiveRecord::Base.connection_handler.clear_active_connections!(:all)
             else
               ActiveRecord::Base.clear_active_connections!
             end


mensfeld:
I'm not sure if we even need this release now due to this PR: https://github.com/karafka/karafka/pull/2233

:thinking:

mensfeld:
Yeah I think we can drop it all together because of the rails wrapper introduced recently

mensfeld:
from the docs: https://guides.rubyonrails.org/threading_and_code_execution.html

>     track which threads are in safe positions for autoloading and reloading
>    enable and disable the Active Record query cache
>    return acquired Active Record connections to the pool
>    constrain internal cache lifetimes

mensfeld:
Going to merge it because I do appreciate the contribution a lot :)

It is however going to be superseded by https://github.com/karafka/karafka/pull/2258

PR #2304, opened 2024-10-07
Report partition downscale as ignored in declarative topics

mensfeld:
continuation of https://github.com/karafka/karafka/pull/2303


catkins (approved):
Love it @mensfeld

PR #2313, opened 2024-10-09
Support detailed exit code on topic commands

mensfeld:
close https://github.com/karafka/karafka/issues/2310

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 2.4.13 (Unreleased)
 - [Enhancement] Make declarative topics return different exit codes on migrable/non-migrable states (0 - no changes, 2 - changes).
+- [Enhancement] Introduce `config.strict_declarative_topics` that should force declaratives on all non-pattern based topics and DLQ topics
 - [Enhancement] Report ignored repartitioning to lower number of partitions in declarative topics.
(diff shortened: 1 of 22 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Karafka Framework Changelog
 
 ## 2.4.13 (Unreleased)
+- [Enhancement] Make declarative topics return different exit codes on migrable/non-mibrable states (0 - no changes, 2 - changes).
```
nijikon, line 4: non-migrable

lib/karafka/cli/topics.rb
```diff
@@ -10,26 +10,51 @@ class Topics < Base
...
-        when 'align'
-          Topics::Align.new.call
-        when 'plan'
-          Topics::Plan.new.call
-        else
-          raise ::ArgumentError, "Invalid topics action: #{action}"
-        end
+        detailed_exit_code = options.fetch(:detailed_exitcode, false)
+
+        changes = case action
```
nijikon, line 36: Can we extract this `case` block to a separate method?

mensfeld: I think it's clear in one method and does not violate rubocop rules. I would only extract the execution because its duplicated.

PR #2331, opened 2024-11-08
support rails 8 in integrations

mensfeld:
close https://github.com/karafka/karafka/issues/2330


spec/integrations/pro/rails/rails80_pristine/just_a_dependency/railtie_setup_spec.rb
```diff
@@ -1,13 +1,14 @@
 # frozen_string_literal: true
 
-# Karafka+Pro should work with Rails 6 using the default setup
+# Karafka+Pro should work with Rails 7 using the default setup
```
nijikon, line 3: Rails 8

nijikon: We can actually skip the version in the comment since it can be determined from the folder name.

PR #2334, opened 2024-11-09
resolve ruby 3.1 deps fallback

mensfeld:
This PR is suppose to resolve issues because of using newer ActiveSupport and friends in Gemfile.lock that are not compatible with Ruby 3.1

We do want to support Ruby 3.1 but we also want to make sure we can run specs on newer versions with newer defaults.

Ref https://github.com/karafka/karafka/pull/2328/files

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,13 @@ jobs:
         run: |
           docker compose up -d || (sleep 5 && docker compose up -d)
 
+      # Newer versions of ActiveSupport and Rails do not work with Ruby 3.1 anymore.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -89,6 +89,13 @@ jobs:
         run: |
           docker compose up -d || (sleep 5 && docker compose up -d)
 
+      # Newer versions of ActiveSupport and Rails do not work with Ruby 3.1 anymore.
+      # While we use newer by default we do want to resolve older and test, thus we remove
+      # Gemfile and let it resolve to the most compatible version possible
```
nijikon, line 94: Should be Gemfile.lock instead of Gemfile

PR #2339, opened 2024-11-15
Update Datadog metrics listener to avoid creating extra arrays (#2338)

mensfeld:
continuation of https://github.com/karafka/karafka/pull/2338


mensfeld:
all good - the preview1 error is a ci load error that appears once in a while and needs different investigation

PR #2345, opened 2024-11-21
Add redraw method

dorner:
This allows testing code to define its own routes _without_ blowing away the defaults. I've found this helpful when I have overall defaults that I want to keep around, but I want to test my own routing plugin with various routes.

### DIFF
--- a/lib/karafka/routing/builder.rb
+++ b/lib/karafka/routing/builder.rb
@@ -67,6 +67,9 @@ module Karafka
         end
       end
 
+      # Clear out the drawn routes.
+      alias :array_clear :clear
+
       # Clear routes and draw them again with the given block. Helpful for testing purposes.
       # @param block [Proc] block we will evaluate within the builder context
       def redraw(&block)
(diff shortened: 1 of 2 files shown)


lib/karafka/routing/builder.rb
```diff
@@ -67,6 +67,19 @@ def draw(&block)
         end
       end
 
+      # Clear out the drawn routes.
+      alias array_clear clear
```
mensfeld, line 71: why 2 names?

mensfeld (changes requested):

lib/karafka/routing/builder.rb
mensfeld: Ah I get it now!

Since it's used internally only could we maybe make it private not to "expose" a false API?

mensfeld (changes requested):

dorner:
Hmm... I haven't found a good way to do this. `private alias x y` seems to be a syntax error, and the other approaches have caused other problems. Do you know of a way to do this?

dorner:
Never mind, ChatGPT to the rescue :)

mensfeld (approved):

PR #2362, opened 2024-12-09
provide the wrap api

mensfeld:
This PR introduces a consumer level `#wrap` API that wraps the whole Karafka consumer execution flow.

In general karafka consumer execution (whether it's a consume, idle or anything else) has two parts:

- user part (user code)
- framework code (our code)

The issue we want to solve with this API is being able to use user set context (like the same transactional producer) for both user code and framework code. Since there is no way now to "assign" transactional producer for the execution time, it cannot be assigned and release after everything is done.

This API fixes this by providing the `#wrap` on consumer that needs to be yielded and that will run the whole job consumer context including framework management stuff that happens post user code execution.

It is mainly needed to normalize the transactional flows in such a way that ensures that post-execution framework managed offset marking is done with the same resources as user transactional stuff.

FYI I also checked the yielding performance on the 4 levels of yielding:

```ruby
puts Benchmark.measure { 100000.times { d{c{b{a{}}}} } }
  0.019726   0.000287   0.020013 (  0.020016)
```

given the fact that it is a per job/batch operation and not a high-frequency execution, this impact is negligible.

### DIFF
--- a/lib/karafka/processing/worker.rb
+++ b/lib/karafka/processing/worker.rb
@@ -19,7 +19,8 @@ module Karafka
     class Worker
       include Helpers::Async
       include Helpers::ConfigImporter.new(
-        worker_job_call_wrapper: %i[internal processing worker_job_call_wrapper]
+        worker_job_call_wrapper: %i[internal processing worker_job_call_wrapper],
+        monitor: %i[monitor]
       )
 
       # @return [String] id of this worker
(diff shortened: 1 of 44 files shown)


lib/karafka/processing/worker.rb
```diff
@@ -53,28 +53,30 @@ def process
         instrument_details = { caller: self, job: job, jobs_queue: @jobs_queue }
 
         if job
-          Karafka.monitor.instrument('worker.process', instrument_details)
+          job.wrap do
+            Karafka.monitor.instrument('worker.process', instrument_details)
```
nijikon, line 57: Can we also import `monitor` here and refer to it only with `monitor` instead of `Karafka.monitor`?

PR #2480, opened 2025-03-12
workflow hardening

mensfeld:
I'm hardening the workflows to introduce self-hosted runners future usage


mensfeld:
FYI I did the numbers @nijikon and as of now it is still not worth moving to self-hosted but those changes anyhow will benefit the security.

Issue #2573, opened 2025-05-02
Add custom signal for rolling restarts with distinct exit code


mensfeld:
## Problem

When using Karafka with systemd in production, there's no way to distinguish between normal shutdowns and restart requests during rolling deployments. This makes implementing zero-downtime deployments challenging.

## Proposal

Add support for a new signal (e.g., `USR2`) that would:
1. Initiate a clean shutdown (like `TERM`)
2. Exit with a specific exit code (e.g., `4`) indicating a controlled restart request

## Benefits

- Enable clean rolling restarts with systemd
- Prevent false alerts during planned deployments
- Allow systemd to be configured to only restart on this specific exit code using `RestartForceExitStatus=4`

## Example systemd config

```
[Service]
ExecStart=/bin/bash -c 'bundle exec karafka server'
Restart=on-failure
RestartForceExitStatus=4
```

This would allow operators to trigger graceful rolling restarts while ensuring all in-process messages are properly handled before restarting.

PR #2575, opened 2025-05-02
support custom exit code on graceful restarts

mensfeld:
close https://github.com/karafka/karafka/issues/2573


mensfeld:
specs exits need to be mocked

PR #2576, opened 2025-05-03
fix past dispatch grace period

mensfeld:
This PR ensures that messages scheduled in close past are not affected by the skip on grace period.

This affects only messages scheduled in the close past but since we allow that, they should be dispatched as any others.


mensfeld:
superseeded by https://github.com/karafka/karafka/pull/2579

Issue #2578, opened 2025-05-03
Max epoch tracking for early cleanup causes messages to be skipped until reload


mensfeld:

mensfeld:
superseeded by https://github.com/karafka/karafka/pull/2580

PR #2579, opened 2025-05-03
Fix max epoch tracking for early cleanup causes messages to be skipped until reload

mensfeld:
close https://github.com/karafka/karafka/issues/2578


mensfeld:
superseeded by https://github.com/karafka/karafka/pull/2580

PR #3204, opened 2026-06-11
set 2.6.0 as a target

mensfeld:


mensfeld:
Note to self: CI will fail because of namespace conflicts but in order to solve it we need one more round of beta releases between karafka and web due to namespace changes.

PR #3251, opened 2026-06-24
Add Kubernetes::ReadinessListener

daniel-niknam-buildkite:
## Add `Kubernetes::ReadinessListener` (draft, for discussion)

Opening this as a **draft** following our chat. Keen to get your steer on the shape before polishing.

### What it does

`Kubernetes::ReadinessListener` serves a readiness probe on its own HTTP port and reports healthy once the consumer has started polling, so a pod is marked Ready when it is consuming. It is meant to be used together with `LivenessListener`: point a `livenessProbe` at the liveness listener (restart on hang) and a `startupProbe` / `readinessProbe` at the readiness listener (gate the rollout until the consumer is polling).

Readiness is reported once **every active subscription group has emitted `connection.listener.fetch_loop` at least once**, and reports **not-ready again once the process is shutting down or quieting** (`Karafka::App.done?`), so Kubernetes can remove a draining pod from the Service endpoints before it exits. The `LivenessListener` HTTP server binds on `app_running`, which is before the consumer has joined its groups and started polling, so on its own it does not distinguish "process up" from "consumer polling"; the readiness listener fills that role.

```ruby
# liveness: restart on hang
Karafka.monitor.subscribe(
  Karafka::Instrumentation::Vendors::Kubernetes::LivenessListener.new(port: 3000)
)
# readiness: report Ready once the consumer is polling
Karafka.monitor.subscribe(
  Karafka::Instrumentation::Vendors::Kubernetes::ReadinessListener.new(port: 3001)
)
```

```yaml
livenessProbe:   { httpGet: { path: /, port: 3000 } }
startupProbe:    { httpGet: { path: /, port: 3001 } }
readinessProbe:  { httpGet: { path: /, port: 3001 } }
```

### Design notes / decisions I'd love your opinion on

- **All subscription groups, not just the first tick.** A process runs one listener thread per subscription group; gating on the first tick of any group would report Ready while other groups are still joining. The expected set is `Karafka::App.subscription_groups` (already post CLI `--include`/`--exclude` filtered, and its ids match the `fetch_loop` event ids), so it matches the groups the process runs.
- **Latches the "all groups polled" condition, but re-checks draining each request.** The all-groups-polled condition only goes false to true, so it does not flap during a rebalance. Overall readiness, though, also requires the process not be `done?` (`quieting`/`stopping`/etc.), and that is re-evaluated on every probe, so a draining pod reports not-ready and can be pulled from the Service endpoints.
- **Fallback.** If the expected set cannot be determined (routes not drawn, or an error), it reports healthy after any one group has polled, so a discovery failure cannot keep a pod from ever becoming Ready.
- **Tracks by subscription-group id** (the liveness listener tracks polling per fiber), since readiness is a per-group "has it started polling" question.

Open questions for you:
1. Is a separate `ReadinessListener` the shape you'd want, or would you prefer this as a mode on the existing listener?
2. Should the readiness definition be configurable (e.g. "first poll of any group" vs "all groups"), or is "all active groups" a good fixed default?
3. Anything about the `status_body` shape / naming you'd want aligned with the liveness listener.

### Notes

- Lives under `lib/karafka/instrumentation/vendors/kubernetes/` (OSS), subclasses `BaseListener`.
- Unit spec added (gate / latch / fallback / `NotificationsChecker` / thread-safety). The HTTP-server lifecycle is left to the integration suite, matching `liveness_listener_spec`/`base_listener_spec`.
- CHANGELOG entry under 2.6.0 (Unreleased).
- We run a subclass of `LivenessListener` doing this internally today; this PR proposes it as a separate readiness listener.


mensfeld:
Great work. I may need few more days for this but it's on my radar.

mensfeld:
superseeded by https://github.com/karafka/karafka/pull/3333 with the original contributions intact

PR #3257, opened 2026-06-27
Dispatch recurring tasks schedule snapshot synchronously for durability

mensfeld:
The schedule snapshot persists each task's previous_time after it runs, but it was dispatched with produce_async. A crash after a task executed but before the async snapshot was delivered would lose the previous_time update, and the next replay would restore the older previous_time and run the task again (double execution).

The schedule snapshot is now dispatched synchronously, so previous_time is durably stored before the executor proceeds. Commands and execution logs stay async. Snapshots only fire when a task actually runs or changes, so the sync cost is negligible.

A full crash reproduction needs an ungraceful kill (a graceful shutdown flushes the async buffer), so the integration spec instead asserts the durability guarantee directly via WaterDrop's produced-sync/produced-async instrumentation: the schedule snapshot must go out synchronously. It fails without the fix (the snapshot is produced async).


mensfeld:
Not worth it after rethinking.

PR #3298, opened 2026-08-03
v2.6.0.rc2

mensfeld:


mensfeld:
The CI here is expected to fail as it runs atm against the web rc1 which uses the removed APIs.

PR #3333, opened 2026-08-25
Readiness listener

mensfeld:
Questions from the original PR:

Open questions for you:

- Is a separate ReadinessListener the shape you'd want, or would you prefer this as a mode on the existing listener?
- Should the readiness definition be configurable (e.g. "first poll of any group" vs "all groups"), or is "all active groups" a good fixed default?
- Anything about the status_body shape / naming you'd want aligned with the liveness listener.


mensfeld:
This one is good to go once green.

